PR #14, opened 2024-06-21
Listen websocket on same port, allow direct connections and more helpful logging

tornadocontrib:
**Description**

Listen websocket on same port, allow direct connections and more helpful logging

**Tests**

**Additional context**

**Metadata**

- Fixes #[Link to Issue]


tornadocontrib:
cc @felipe-op could you check?

jelias2:
Closing out due to inactivity. At this time do not see a need for this feature, and would like to keep our scope of maintenance to a minimal set of functionality.

PR #15, opened 2024-06-24
feat: Ignore block height zero response within UpdateBackend

jelias2:
**Description**
* This PR updates `proxyd` to handle block height zero responses. Previously, a block height of zero might have been returned if `eth_getBlockByNumber` encountered an error, which would have caused the backend to be banned immediately. This PR modifies the logic so that a block height of zero or a network error will not overwrite the backend state, thus preventing the immediate banning of the node.

  The node can still be banned if it reaches the maximum network errors threshold.

**Tests**
* Added tests to verify that backend state will not be overwritten, and that a node will not be banned if an error and a block height of zero are received.
* Added tests to ensure that consensus can still advance if the latest block is not responding correctly.

**Additional Context**
* Adding the new functionality to the consensus poller caused a specific test in `consensus_test.go` to break: "ban backend if error rate is too high".

  The test broke because the network sliding errors window did not reach the ban threshold within the current test case. It was found that the network sliding errors window was not reset between test cases, and the "ban backend if error rate is too high" test was relying on a previously instantiated sliding window state. I added code to reset the sliding window between test cases to ensure it now passes. 
  
  * Getters for additional backend state and clearing sliding windows have been added.

**Metadata**

- Fixes [Issue-139](https://github.com/ethereum-optimism/devinfra-pod/issues/139)


.circleci/continue_config.yml
```diff
@@ -462,6 +462,8 @@ workflows:
           docker_name: op-conductor-mon
           docker_tags: <<pipeline.git.revision>>,<<pipeline.git.branch>>
           docker_context: .
+          requires:
```
felipe-op, line 465: is this supposed to be part of this change?

proxyd/consensus_poller.go
```diff
@@ -318,18 +336,22 @@ func (cp *ConsensusPoller) UpdateBackend(ctx context.Context, be *Backend) {
 	}
 
 	latestBlockNumber, latestBlockHash, err := cp.fetchBlock(ctx, be, "latest")
-	if err != nil {
-		log.Warn("error updating backend - latest block", "name", be.Name, "err", err)
+	if err != nil || latestBlockNumber == 0 {
```
felipe-op: when block is 0, we should increment `networkErrorsSlidingWindow`, also this var should be renamed to `intermittentErrorsSlidingWindow`

proxyd/consensus_poller.go
```diff
@@ -318,18 +336,22 @@ func (cp *ConsensusPoller) UpdateBackend(ctx context.Context, be *Backend) {
...
+		log.Warn("error updating backend - latest block will not be updated", "name", be.Name, "err", err)
```
felipe-op, line 346: we should also return from here, no point in trying to update the other fields after some have already failed

proxyd/consensus_poller.go
```diff
@@ -318,18 +336,22 @@ func (cp *ConsensusPoller) UpdateBackend(ctx context.Context, be *Backend) {
...
+		latestBlockNumber = bs.latestBlockNumber
+		latestBlockHash = bs.latestBlockHash
 	}
 
 	safeBlockNumber, _, err := cp.fetchBlock(ctx, be, "safe")
```
felipe-op, line 355: safe could potentially be zero as well

proxyd/consensus_poller.go
```diff
@@ -318,18 +336,22 @@ func (cp *ConsensusPoller) UpdateBackend(ctx context.Context, be *Backend) {
...
 	if err != nil {
-		log.Warn("error updating backend - safe block", "name", be.Name, "err", err)
+		log.Warn("error updating backend - safe block will not be updated", "name", be.Name, "err", err)
+		safeBlockNumber = bs.safeBlockNumber
 	}
 
 	finalizedBlockNumber, _, err := cp.fetchBlock(ctx, be, "finalized")
```
felipe-op, line 367: finalized could potentially be zero as well

felipe-op:
instead of banning the backend on unexpected block tags, let's just increment the error sliding window here:
https://github.com/ethereum-optimism/infra/blob/75b33efb4eea32d585a7a2fe66882ef5d913219e/proxyd/consensus_poller.go#L396

felipe-op (changes requested):

proxyd/consensus_poller.go
felipe-op: any error on this method should prematurely return and not update the backend state

if the backend causes too much errors it will automatically breach the threshold of the intermittent errors and we dont need to explicit ban it

.circleci/continue_config.yml
jelias2: Prior to this change when the branch was up for PR, it was executing the op-conductor-mon release worklow and failing. 

This changed stoped that from executing, I'm happy to open up this smaller update within a new PR

proxyd/consensus_poller.go
jelias2: Happy to rename the var. 

Just to clarify on any non 200 the networkErrorsSlidingWindow is incremented within the fetchBlock function already

This update would mean the sliding window would then be incremented twice if a blockNumber 0 came in or a non-200. 

Please confirm if this is desired.  

Happy to brainstorm other solutions too if we need a different behavior, I do have a wip branch with a sliding window just for the the block height zeros

proxyd/consensus_poller.go
jelias2: Sure I can return from here. 

That also means we can remove setting the backend state variable because it will never be executed with consensus.SetBackendState

I implemented it this way so that the safe and finalized blocks could still advance if we wanted that

proxyd/consensus_poller.go
jelias2: Agreed I can update this so that if its zero too we return

proxyd/consensus_poller.go
jelias2: Agreed I can update this so that if its zero too we return

proxyd/consensus_poller.go
felipe-op: but that would return an `err` right?

we can avoid increment dupes this way:

```
if err != nil {
 // handle error and return
} 

// the statement below will only be reached in case there is no error
if blockHeight == 0 {
  // increment the window and return
}  
```

.circleci/continue_config.yml
felipe-op: no needed, just want to confirm/clarify

proxyd/consensus_poller.go
felipe-op: I'm afraid of leaving edgy cases if we do partial updates, so lets avoid it entirely for now

it will be a full update or nothing

proxyd/consensus_poller.go
```diff
@@ -306,30 +323,56 @@ func (cp *ConsensusPoller) UpdateBackend(ctx context.Context, be *Backend) {
...
 	if !be.skipPeerCountCheck {
 		peerCount, err = cp.getPeerCount(ctx, be)
 		if err != nil {
 			log.Warn("error updating backend peer count", "name", be.Name, "err", err)
+			return
+		}
+		if peerCount == 0 {
+			log.Warn("peer count responded with 200 and 0 peers", "name", be.Name)
+			return
 		}
```
jelias2: Just to confirm this is the expected behavior for peer count zero?

felipe-op: we can remove this check

```
		if peerCount == 0 {
			log.Warn("peer count responded with 200 and 0 peers", "name", be.Name)
			return
		}
```

proxyd/consensus_poller.go
```diff
@@ -306,30 +323,56 @@ func (cp *ConsensusPoller) UpdateBackend(ctx context.Context, be *Backend) {
...
+		log.Warn("error backend responded a 200 with blockheight 0 for latest block", "name", be.Name)
+		be.intermittentErrorsSlidingWindow.Incr()
+		return
 	}
 
 	safeBlockNumber, _, err := cp.fetchBlock(ctx, be, "safe")
 	if err != nil {
-		log.Warn("error updating backend - safe block", "name", be.Name, "err", err)
+		log.Warn("error updating backend - safe block will not be updated", "name", be.Name, "err", err)
+		safeBlockNumber = bs.safeBlockNumber
```
felipe-op: return from here

felipe-op: why do we have this assignment?
feels buggy
```
  safeBlockNumber = bs.safeBlockNumber
```

proxyd/consensus_poller.go
```diff
@@ -306,30 +323,56 @@ func (cp *ConsensusPoller) UpdateBackend(ctx context.Context, be *Backend) {
...
+		be.intermittentErrorsSlidingWindow.Incr()
+		return
 	}
 
 	finalizedBlockNumber, _, err := cp.fetchBlock(ctx, be, "finalized")
 	if err != nil {
-		log.Warn("error updating backend - finalized block", "name", be.Name, "err", err)
+		log.Warn("error updating backend - finalized block will not be updated", "name", be.Name, "err", err)
+		finalizedBlockNumber = bs.finalizedBlockNumber
+	}
```
felipe-op, line 371: return from here

proxyd/consensus_poller.go
jelias2: Good find, left this by accident

proxyd/consensus_poller.go
jelias2: Thanks for finding this

felipe-op:
lgtm
please squash before merging

felipe-op (approved):

PR #17, opened 2024-06-26
Multi call test use fanout for all batched

jelias2:
**Description**

**Tests**

**Additional context**

**Metadata**

- Fixes #[Link to Issue]


proxyd/backend.go
```diff
@@ -739,6 +750,15 @@ func (bg *BackendGroup) Primaries() []*Backend {
 	return primaries
 }
 
+func isValidMultiCallTx(rpcReqs []*RPCReq) bool {
+	for _, r := range rpcReqs {
+		if r.Method == "eth_sendRawTransaction" {
+			return true && len(rpcReqs) == 1
```
felipe-op: seems like `&& len(rpcReqs) == 1` this should be in the line above

proxyd/backend.go
```diff
@@ -711,11 +711,12 @@ func sortBatchRPCResponse(req []*RPCReq, res []*RPCRes) {
...
-	Backends         []*Backend
-	WeightedRouting  bool
-	Consensus        *ConsensusPoller
-	FallbackBackends map[string]bool
+	Name              string
+	Backends          []*Backend
+	WeightedRouting   bool
+	Consensus         *ConsensusPoller
+	FallbackBackends  map[string]bool
+	MulticallBackends map[string]bool
```
felipe-op: how this will be configured?

I thought we would just have a new config option in the backend group with three potential values. This should replace the `consensus_aware` boolean to something like `routing_strategy` which can be `fallback`, `consensus` or `fanout`. 

Configuring a bg as a fallback should just use the same list of backends we already have.

felipe-op (changes requested):

proxyd/backend.go
felipe-op: also you may just want to use the `IsBatched` function

jelias2: Sure I can update this

proxyd/backend.go
jelias2: > how this will be configured?

The Multicall backends map is currently populated from the multicall array of backends in the toml [config](https://github.com/ethereum-optimism/infra/pull/17/files#diff-4536a43e4347524e6136ba0185787a1e5317ca26b6ab7cf5a7aaa30611583ebeR30), this was just a starting point. 

If i understand correctly, we will plan to configure routing strategy option as multicall for a new backend group, and then we would only use the multicall backend group for eth_sendRawTranscations RPCs, all other rpcs would be routed to the normal backend group.

I can update the code for this change

PR #26, opened 2024-07-15
feat: add multi call routing strategy

jelias2:
**Description**
  * Updated proxyd config to accept a routing strategy paramter instead of consenus_aware boolean
  * Added multicall routing strategies
  * Refactored backendGroup.Forward to handle multiple types of routing strategies
  * Background.ForwardToBackend now will return results via a channel

**Tests**
* Unit Testing in multi call unit testing to ensure that first successfully application level response will be returned to caller, network level errors will be logged and ignored.
* Tested in a local environment with varied levels of situations using invalid RPCs, 403 for eth_sendRpcTransaction, and ore

**Additional context**
* Breaks existing configurations as conseusus_aware = true, is no longer a valid config value. New value is routing_strategy = consensus_aware

**Metadata**

- Fixes #[166](https://github.com/ethereum-optimism/devinfra-pod/issues/166)

### DIFF
--- a/proxyd/config.go
+++ b/proxyd/config.go
@@ -112,6 +112,7 @@ type RoutingStrategy string
 const (
 	ConsensusAwareRoutingStrategy RoutingStrategy = "consensus_aware"
 	MulticallRoutingStrategy      RoutingStrategy = "multicall"
+	FallbackRoutingStrategy       RoutingStrategy = "fallback"
 )
 
 type BackendGroupConfig struct {
--- a/proxyd/config.go
+++ b/proxyd/config.go
@@ -136,8 +136,6 @@ type BackendGroupConfig struct {
 	ConsensusHARedis             RedisConfig  `toml:"consensus_ha_redis"`
 
 	Fallbacks []string `toml:"fallbacks"`
-
-	MulticallBackends []string `toml:"multicalls"`
 }
 
 type BackendGroupsConfig map[string]*BackendGroupConfig
--- a/proxyd/config.go
+++ b/proxyd/config.go
@@ -107,12 +107,20 @@ type BackendConfig struct {
 
 type BackendsConfig map[string]*BackendConfig
 
+type RoutingStrategy string
+
+const (
+	ConsensusAwareRoutingStrategy RoutingStrategy = "consensus_aware"
+	MulticallRoutingStrategy      RoutingStrategy = "multicall"
+)
+
 type BackendGroupConfig struct {
 	Backends []string `toml:"backends"`
 
 	WeightedRouting bool `toml:"weighted_routing"`
 
-	ConsensusAware          bool         `toml:"consensus_aware"`
+	RoutingStrategy RoutingStrategy `toml:"routing_strategy"`
+
 	ConsensusAsyncHandler   string       `toml:"consensus_handler"`
 	ConsensusPollerInterval TOMLDuration `toml:"consensus_poller_interval"`
 
@@ -128,6 +136,8 @@ type BackendGroupConfig struct {
 	ConsensusHARedis             RedisConfig  `toml:"consensus_ha_redis"`
 
 	Fallbacks []string `toml:"fallbacks"`
+
+	MulticallBackends []string `toml:"multicalls"`
 }
 
 type BackendGroupsConfig map[string]*BackendGroupConfig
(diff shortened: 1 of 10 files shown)


proxyd/proxyd.go
```diff
@@ -416,6 +417,10 @@ func Start(config *Config) (*Server, func(), error) {
 			if bgcfg.ConsensusHA {
 				tracker.(*RedisConsensusTracker).Init()
 			}
+		} else if bgcfg.RoutingStrategy == MulticallRoutingStrategy {
+			log.Info("configuring routing multicall strategy for backend_group", "name", bgName, "routing_strategy", bgcfg.RoutingStrategy)
+		} else {
+			log.Info("no routing strategy defined for backend_group using default round robin", "name", bgName, "routing_strategy", bgcfg.RoutingStrategy)
```
felipe-op: the default should be fallback, not round robin

proxyd/integration_tests/testdata/multicall.toml
```diff
@@ -0,0 +1,33 @@
...
+[backends.node2]
+rpc_url = "$NODE2_URL"
+
+[backends.node3]
+rpc_url = "$NODE3_URL"
+
+[backend_groups]
+[backend_groups.node]
+backends = ["node1", "node2", "node3"]
+routing_strategy = "multicall"
```
felipe-op, line 23: are those two configs necessary?
```
routing_strategy = "multicall"
multicall_mode = true
```

what is the difference between them?

proxyd/config.go
```diff
@@ -128,6 +136,8 @@ type BackendGroupConfig struct {
 	ConsensusHARedis             RedisConfig  `toml:"consensus_ha_redis"`
 
 	Fallbacks []string `toml:"fallbacks"`
+
+	MulticallBackends []string `toml:"multicalls"`
```
felipe-op: didn't we dropped this?

once the bg is configured to multicall, all backends in the group should be multicall backends

felipe-op (changes requested):

proxyd/config.go
jelias2: Correct, its dropped. This is unused and will be removed

proxyd/integration_tests/testdata/multicall.toml
jelias2: mutlicall_mode is not used and will be removed. It was used in the first draft implementation

proxyd/proxyd.go
jelias2: Fallback is where we sequentially request he backends one at a time for a response with no consensus. This is what the default is routing strategy is.  I called this round robin in the log, I can update it to fallback.

felipe-op: Can we have a explicit "fallback" option with its const as well?
Config examples should be updated stating this new routing_strategy config, with default to fallback.

jelias2: Sure, we can have a config option that is explicitly is routing_strategy == fallback, and lack of the the routing_strategy will default to the fallback strategy. 

Happy to add that. 


I'd also propose that we rename the config "fallbacks" which is used for consensus routing strategy to something like consensus_fallbacks to reduce confusion between these configs. Happy to add that in a follow up ticket

felipe-op: One `fallback` is a value for "routing_strategy" and the other is the key that holds the list of fallback backends that is meaningful only in `consensus` mode, right?

If so, its ok to keep the same name.

Also, let's make this config backwards compatible:
* Mark the legacy `consensus_aware` as deprecated with some comments indicating how it should be specified now
* If `consensus_aware` is set to true, it should implicit set "routing_strategy" to `consensus` and print a INFO log at initialization
* If both `consensus_aware` and `routing_strategy` are populated at same time, it should fail at initialization with a message indicating they are mutual exclusive and "routing_strategy" is the preferred way moving forward

jelias2: Do you want an explicit routing strategy value for fallback, in addition to fallback being the default behavior if none is provided?

felipe-op: yes

the config "language" definition should be
routing_strategy = fallback | consensus_aware | multicall

with fallback being the default in cause its omitted

the [if/else](https://github.com/ethereum-optimism/infra/pull/26/files#diff-01f9fee1156d98d875869e58c791a9602554a9e06932e224dcd0567e4cdad7b3R430) used to pick the right strategy *should not* handle the fallback case as a catch-all else

jelias2: Provided a function to handle this logic, the catch-all will fail

felipe-op (approved):

PR #27, opened 2024-07-19
fix: context ends in WS after HTTP upgrade

mkykadir:
**Description**

After HTTP connection upgrades to WS connection, context from HTTP ends and this prevents context to be utilized where it is required to continue, such as accessing Redis operations within WS.

**Additional context**

Cannot access to Redis within WS since context ends


jelias2:
Closing due to staleness, feel free to open with a rebased pr

PR #29, opened 2024-07-19
feat: report method name in WS metrics

mkykadir:
**Description**

Access method name in WS context so metrics can be reported with method name

### DIFF
--- a/proxyd/backend.go
+++ b/proxyd/backend.go
@@ -976,7 +976,6 @@ type WSProxier struct {
 	methodWhitelist *StringSet
 	readTimeout     time.Duration
 	writeTimeout    time.Duration
-	reqIdToMethod   map[string]string
 }
 
 func NewWSProxier(backend *Backend, clientConn, backendConn *websocket.Conn, methodWhitelist *StringSet) *WSProxier {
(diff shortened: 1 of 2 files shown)


jelias2:
Hi Mkykadir,

Thanks for the contribution!

Its a clever solution, however I am concerned about a race condition occurring with concurrent map access w.reqIdToMethod if the proxyd was under heavy ws traffic. 

I'm curious if you would be able to find a solution that can avoid this? 

Also please include test cases if possible

jelias2 (changes requested):
In the PRs current state with concurrent map access, I believe these concerns must be addressed before merging

mkykadir:
@jelias2 nice catches actually I missed a crucial point as well:
* reqId within context is determined on WS connection and shouldn't be used for all following ws messages
* now each ws connection will have its own concurrent map
* and following message ids will be used for this
* after its ok for you I'll go with tests as well

proxyd/string_map.go
```diff
@@ -0,0 +1,56 @@
+package proxyd
```
jelias2, line 1: Some tests for this struct to validate concurrency would nice

proxyd/backend.go
```diff
@@ -1105,7 +1108,14 @@ func (w *WSProxier) backendPump(ctx context.Context, errC chan error) {
 			continue
 		}
 
+		method := MethodUnknown
 		res, err := w.parseBackendMsg(msg)
+		if res != nil {
+			if value, has := idToMethod.GetAndRemove(string(res.ID)); has {
```
jelias2, line 1114: Will res.ID be unique? Typically is determined by the request ID with in the rpc, I don't believe its unique among many requests?

PR #31, opened 2024-07-23
Option to disable shuffle

cody-wang-cb:
**Description**
In consensus aware mode, it currently always does shuffling/round robin among the healthy nodes. However it doesn't quite fit with our use case given that we'd like to always route the request to the main node if the main node is healthy. Because of this we have to run the vanilla settings and can't leverage the health checks consensus aware mode provides.

Thus, I'd like to add an option to disable shuffling so that the main node is always attempted first unless it's unhealthy/banned in consensus aware mode.

**Tests**
Added an integration test to test `disable_shuffle: true`.

**Additional context**
N/A

**Metadata**

- Fixes #[Link to Issue]

### DIFF
--- a/proxyd/config.go
+++ b/proxyd/config.go
@@ -151,6 +151,7 @@ type BackendGroupConfig struct {
 	Backends []string `toml:"backends"`
 
 	WeightedRouting bool `toml:"weighted_routing"`
+	DisableShuffle  bool `toml:"disable_shuffle"`
 
 	RoutingStrategy RoutingStrategy `toml:"routing_strategy"`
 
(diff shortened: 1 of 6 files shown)


proxyd/config.go
```diff
@@ -151,6 +151,7 @@ type BackendGroupConfig struct {
 	Backends []string `toml:"backends"`
 
 	WeightedRouting bool `toml:"weighted_routing"`
+	DisableShuffle  bool `toml:"disable_shuffle"`
```
cody-wang-cb, line 154: Maybe using `shuffle` is a better code style but I think everything needs to be by default `false`?

cody-wang-cb:
Close this PR as we can just turn on `weightedRouting` and set the weight of one node to be very high instead

PR #38, opened 2024-08-01
Add healthcheck ordering to simple failover mode

cody-wang-cb:
**Description**
Currently for the vanilla setting (simple failover), it doesn't do any health checks. However it still makes sense to order the endpoints in a way such that the ones that are experiencing high error/high latency should be deprioritized. 

**Tests**
Added a unit test to prove that when an endpoint is experiencing high errors, it would reorder the endpoints so that the good endpoint would be put in front of the bad one.

**Additional context**

**Metadata**

- Fixes #[Link to Issue]

### DIFF
--- a/proxyd/integration_tests/failover_test.go
+++ b/proxyd/integration_tests/failover_test.go
@@ -87,7 +87,7 @@ func TestFailover(t *testing.T) {
 		})
 	}
 
-	// the bad endpoint has had 10 requests with 8 error (3xx/4xx/5xx) responses, it should be marked as unhealthy
+	// the bad endpoint has had 10 requests with 8 error (3xx/4xx/5xx) responses, it should be marked as unhealthy and deprioritized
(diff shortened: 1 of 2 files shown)


proxyd/integration_tests/failover_test.go
```diff
@@ -87,6 +87,38 @@ func TestFailover(t *testing.T) {
...
+
+	require.NoError(t, os.Setenv("GOOD_BACKEND_RPC_URL", goodBackend.URL()))
+	require.NoError(t, os.Setenv("BAD_BACKEND_RPC_URL", badBackend.URL()))
+
+	config := ReadConfig("failover")
+	client := NewProxydClient("http://127.0.0.1:8545")
+	_, shutdown, err := proxyd.Start(config)
+	require.NoError(t, err)
+	defer shutdown()
+
```
cody-wang-cb, line 135: Moving the tests below to an individual test suite so resetting the proxyd

felipe-op (approved):

PR #41, opened 2024-08-13
Use not healthy as error msg

cody-wang-cb:
**Description**
Currently when the backends all fail, it return `no backend is available`, which is quite unclear for the end user. Making a change here to use `no backend is currently healthy to serve traffic` instead which is a lot more clear.

**Tests**
Fixed the impacted tests

**Additional context**

**Metadata**

- Fixes #[Link to Issue]

### DIFF
--- a/proxyd/backend.go
+++ b/proxyd/backend.go
@@ -1451,7 +1451,7 @@ func (bg *BackendGroup) ForwardRequestToBackendGroup(
 	return &BackendGroupRPCResponse{
 		RPCRes:   nil,
 		ServedBy: "",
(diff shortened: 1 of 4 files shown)


jelias2 (approved):
I agree that the returned message could be improved to include more info, so I'll approve the PR. 

However I am hesitant that this message implies that the backed is not healthy, where as  the requests could fail for additional reasons besides just backend health. Thus this error may be slightly misleading. 

Would like to get another set of eyes before merging

felipe-op (approved):

PR #42, opened 2024-08-14
op-txproxy: external validating proxy for conditional transactions

hamdiallam:
**Description**
A passthrough proxy service for some endpoint prior to reaching the sequencer. Currently just the `sendRawTranscationConditional` endpoint. This proxy implements stateless validation as DDoS prevention for this new endpoint
- Authentication: [flashbot like auth](https://docs.flashbots.net/flashbots-auction/advanced/rpc-endpoint#authentication). Under the `X-Optimism-Signature` header
- Rate Limiting
- Metrics
- Enable/Disable Conditional Requests
- 4337 Entrypoint tx target restriction
- Basic conditional validation

DevX is responsible for maintenance of this service & any new changes that may be required with usage. Ideally working closely with DevInfra on the proxyd integration & setup
- The passthrough proxy will rely on the proxyd to broadcast the conditional tx to all replicas as these txs are not p2p gossiped.

**Tests**
The sequencer endpoint has it's own [e2e test](https://github.com/ethereum-optimism/optimism/pull/11671) in the monorepo and tests in op-geth

The PR includes unit tests for this passthrough endpoint. Testing authentication & entrypoint validation

This endpoint can reject traffic behind a flag. Plan being to enable on testnet and ensure the endpoint & dashboards are working correctly with a live bundler prior to enabling on mainnet

**Additional Context**
Requires the PR for the sequencer endpoint to land in op-geth.
* [x] ethereum-optimism/op-geth#330

**Metadata**
Re-opening of ethereum-optimism/optimism#11223. This repo is better suited for this service as supporting infrastructure


op-txproxy/go.mod
```diff
@@ -0,0 +1,93 @@
...
+	golang.org/x/tools v0.23.0 // indirect
+	google.golang.org/protobuf v1.34.1 // indirect
+	gopkg.in/natefinch/lumberjack.v2 v2.0.0 // indirect
+	gopkg.in/yaml.v3 v3.0.1 // indirect
+	rsc.io/tmplfunc v0.0.3 // indirect
+)
+
+//replace github.com/ethereum/go-ethereum => github.com/ethereum-optimism/op-geth v1.101315.3-rc.1
+
+replace github.com/ethereum/go-ethereum => ../../op-geth
```
hamdiallam: Will be commented out and replaced up when dependent PR is merged in

jelias2 (commented):
Hi @hamdiallam would you be able to include a README describing op and how it would fit into our infrastructure?

I'd just want to make sure this fits within the context of the infra repo as infra team is mostly likely the people who will may have to maintain this in the long run

hamdiallam:
> Hi @hamdiallam would you be able to include a README describing op and how it would fit into our infrastructure?
> 
> I'd just want to make sure this fits within the context of the infra repo as infra team is mostly likely the people who will may have to maintain this in the long run

yes I definitely get a README added here

hamdiallam:
@jelias2 added a readme

op-txproxy/conditional_txs.go
```diff
@@ -0,0 +1,143 @@
...
+func NewConditionalTxService(ctx context.Context, log log.Logger, m metrics.Factory, cfg *CLIConfig) (*ConditionalTxService, error) {
+	rpc, err := client.NewRPC(ctx, log, cfg.SendRawTransactionConditionalBackend)
+	if err != nil {
+		return nil, fmt.Errorf("failed to dial backend %s: %w", cfg.SendRawTransactionConditionalBackend, err)
+	}
+
+	rpcMetrics := metrics.MakeRPCClientMetrics("backend", m)
+	backend := client.NewInstrumentedRPC(rpc, &rpcMetrics)
+
+	limiter := rate.NewLimiter(params.TransactionConditionalMaxCost, int(cfg.SendRawTransactionConditionalRateLimit))
```
hamdiallam: Need to update this based on this discussion: https://github.com/ethereum-optimism/op-geth/pull/330#discussion_r1752214483

op-txproxy/auth_handler.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+
+	if r.Body == nil {
+		// edge case from unit tests
+		r.Body = io.NopCloser(bytes.NewBuffer(nil))
+	}
+
+	// The request body can only be read once. Adjust here so it's inspectable
+	var buf bytes.Buffer
+	body := io.TeeReader(r.Body, &buf)
```
anacrolix: You forgot to schedule closing the original Request.Body. This would be very bad in prod.

op-txproxy/auth_handler.go
```diff
@@ -0,0 +1,95 @@
...
+	r.Body = io.NopCloser(&buf)
+
+	// Since this middleware runs prior to the server, we need to manually apply the body limit when reading.
+	bodyBytes, err := io.ReadAll(io.LimitReader(body, int64(defaultBodyLimit)))
```
anacrolix: This will pass only a partial body to the origin (if there's more after the min of defaultBodyLimit and the TextHash). Did you intend this?

op-txproxy/auth_handler.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+
+	if caller != crypto.PubkeyToAddress(*sigPubKey) {
+		http.Error(w, "mismatched recovered signer", http.StatusBadRequest)
+		return
+	}
+
+	// Set the authenticated caller in the context
+	newCtx := context.WithValue(r.Context(), authContextKey{}, &AuthContext{caller})
+	h.next.ServeHTTP(w, r.Clone(newCtx))
```
anacrolix: The Request.Clone is good, but I think it's too conservative here. I think WithContext is fine because you haven't modified the original request (except the Body and that's fine). I'm curious what you think here.

anacrolix (changes requested):
I think there might be a leak.

It's nice to put the auth into the context. Why can't the "next" Handlers just tee the body and do this themselves with a helper?

op-txproxy/auth_handler.go
anacrolix: Ok now I've read more, I wonder if you want to just defer this limit downstream?

If so, read up to the limit, do your auth unpacking, then repack the new Body with io.MultiReader(buf, oldBody). Then downstream gets the rest of the payload and can make its own decisions. Also now that I see you call synchronously into the downstream handler, you can fix the leak above by just deferring r.Body.Close here, particularly if you don't buffer anything that exceeds the body limit you read for auth. This is the way.

op-txproxy/README.md
```diff
@@ -0,0 +1,29 @@
...
+```
+
+## Endpoints
+
+### eth_sendRawTransactionConditional
+
+An outcome of how to integrate this [spec](https://notes.ethereum.org/@yoav/SkaX2lS9j) safely for permissionless 4337 bundler participation. This solution in the design doc [proposal](https://github.com/ethereum-optimism/design-docs/blob/main/ecosystem/sendRawTransactionConditional/proposal.md)
+requires a validating proxy that can be horizontally scaled and pre-emptively reject invalid conditional transaction. The implemented endpoint covers
+these objectives:
+1. **Auth**. any ecdsa key can be used by the caller to authorize themselves. While this endpoint currently implements no auth policies, this has been preemptively put in place to enable this in the future (allowlist, rate limits, etc).
```
tynes: Can you clearly specify how the auth works? The specific header that is used and the format of the data and how it is validated?

op-txproxy/README.md
```diff
@@ -0,0 +1,29 @@
+# op-txproxy
+
+A supplemental passthrough proxy for some execution engine endpoints. This proxy does not forward all rpc traffic and only exposes a specific set of endpoints.
+Operationally, the public ingress proxy should  only re-route requests for these endpoints.
+```
```
tynes: Mermaid could be more maintainable

op-txproxy/README.md
tynes: See https://docs.flashbots.net/flashbots-auction/advanced/rpc-endpoint#authentication

anacrolix (changes requested):

op-txproxy/auth_handler.go
hamdiallam: ah! good catch

op-txproxy/auth_handler.go
hamdiallam: ooh yea I like this suggestion! Thanks

op-txproxy/auth_handler.go
hamdiallam: Ahh I missed `WithContext` on request. Yeah that's much better

hamdiallam:
> It's nice to put the auth into the context. Why can't the "next" Handlers just tee the body and do this themse

Good Q! So because this is using geth lib to construct server, it would require some changes to op-geth in order to allow a json-rpc methods to inspect the request body. See ethereum-optimism/op-geth/pull/352. There were some valid concerns around the required changes.

It seemed simpler to instead create an http middleware as it remains compatible with using geth & actually made this PR a lot more unit-testable

Curious what you think about this

hamdiallam:
> I think there might be a leak.
> 
> It's nice to put the auth into the context. Why can't the "next" Handlers just tee the body and do this themselves with a helper?


Good Q! So because this is using geth lib to construct server, it would require some changes to op-geth in order to allow a json-rpc methods to inspect the request body. See ethereum-optimism/op-geth/pull/352. There were some valid concerns around the required changes.

It seemed simpler to instead create an http middleware as it remains compatible with using geth & actually made this PR a lot more unit-testable

Curious what you think about this

op-txproxy/auth_handler.go
anacrolix: Cool. I didn't approve because there's a leak but I think it will be a great improvement.

anacrolix:
Yeah it made sense after I read a bit further. Cheers

op-txproxy/auth_handler.go
hamdiallam: Updated!

op-txproxy/README.md
hamdiallam: sg! Lmk how it looks now

op-txproxy/README.md
hamdiallam: ah nice, i'll use that

hamdiallam:
@tynes @anacrolix soft bump on the latest changes

op-txproxy/auth_handler.go
```diff
@@ -56,18 +59,17 @@ func (h *authHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 	// Since this middleware runs prior to the server, we need to manually apply the body limit when reading.
-	bodyBytes, err := io.ReadAll(io.LimitReader(body, int64(defaultBodyLimit)))
+	bodyBytes, err := io.ReadAll(io.LimitReader(r.Body, int64(defaultBodyLimit)))
 	if err != nil {
 		http.Error(w, "unable to parse request body", http.StatusInternalServerError)
 		return
 	}
 
+	// The request body can only be read once so place back in the bytes along with whatever else may have been present
+	defer r.Body.Close()
```
anacrolix: I might have been thinking of Response.Body wrt to leaking without Close. I think you can remove this.

op-txproxy/auth_handler.go
```diff
@@ -56,18 +59,17 @@ func (h *authHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	r.Body = io.NopCloser(io.MultiReader(bytes.NewReader(bodyBytes), r.Body))
```
anacrolix: Forward responsibility and the ability to close to the next handler.
```suggestion
    r.Body = interface{
        io.Reader
        io.Closer
    }{
        io.MultiReader(bytes.NewReader(bodyBytes), r.Body)),
        r.Body,
    }
```

anacrolix (approved):

op-txproxy/auth_handler.go
hamdiallam: ah niiice this is sleek

op-txproxy/go.mod
```diff
@@ -0,0 +1,92 @@
+module github.com/ethereum-optimism/optimism/op-txproxy
```
jelias2: Please update the module 
```suggestion
module github.com/ethereum-optimism/infra/op-txproxy
```

jelias2 (commented):
Overall LGTM, two small items

I would recommend adding the op-txproxy to the tag-service file [here](https://github.com/jelias2/infra/blob/88521bef299e8d87eece8b375b3be5214baa2b39/ops/tag-service/tag-tool.py?plain=1#L6-L11) so the repo can publish releases. 

Would you also mind adding an appropraite CODEOWNERS entry in .github for this directory

op-txproxy/go.mod
hamdiallam: ahh nice catch, had this service in the monorepo initially

jelias2 (approved):

PR #56, opened 2024-09-30
Upgrade geth dependency to 1.14.8

palango:
**Description**

We want to use `proxyd` in Celo, but currently the transaction validation of Celo specific transactions fails, because `proxyd` uses geth to validate them.

For that reason we want to use Celo's fork of op-geth, which is based on a later geth version (as is used by the optimsm and op-geth repos).

This PR updates the `go-ethereum` version to 1.14.8 and adapts required logging code.

**Tests**

Tests pass locally.


palango:
@jelias2 could you please have a look at this?

jelias2 (approved):

PR #64, opened 2024-10-17
Multicall rpc error check

cody-wang-cb:
**Description**

Add `bool` to have an option that for multiple success responses from multicall, choose the one that doesn't have rpc error.

**Tests**

Added a unit test

**Additional context**

We observed that there are cases when different execution clients gave different responses, one with rpc success and one with rpc error, if the one with a rpc error returned first it'd return that. However in such case the rpc success one should be returned back to the user

**Metadata**

- Fixes #[Link to Issue]


cody-wang-cb:
@jelias2 @danyalprout PTAL, will add unit tests if this approach makes sense

proxyd/backend.go
```diff
@@ -896,9 +897,11 @@ func (bg *BackendGroup) MulticallRequest(backend *Backend, rpcReqs []*RPCReq, wg
 	}
 }
 
-func (bg *BackendGroup) ProcessMulticallResponses(ch chan *multicallTuple, ctx context.Context) *BackendGroupRPCResponse {
-	var finalResp *BackendGroupRPCResponse
+func (bg *BackendGroup) ProcessMulticallResponses(ch chan *multicallTuple, ctx context.Context, multicallRPCErrorCheck bool) *BackendGroupRPCResponse {
```
danyalprout: Isn't `multicallRPCErrorCheck` a property on `bg`? Probably don't need to pass it through

proxyd/backend.go
```diff
@@ -921,22 +948,42 @@ func (bg *BackendGroup) ProcessMulticallResponses(ch chan *multicallTuple, ctx c
...
+			errResp = resp
 			continue
 		}
+
+		// success or rpc error response
+		if multicallRPCErrorCheck {
+			// check if any of the response is an RPC error
+			var rpcError = false
+			for _, rpcRes := range resp.RPCRes {
+				// check if the ID is already in the success response
```
cody-wang-cb: I think ID should be unique given it's part of the `handleBatchRPC` flow. It's either the incoming message is a batch request (which should have unique IDs otherwise they wouldn't know how to process the response), or a single request.
@jelias2 would be good if you could help to confirm.
cc: @danyalprout

proxyd/backend.go
cody-wang-cb: yup make sense

proxyd/backend.go
jelias2: Theres one key thing is that currently we don't allow batch requests in multicall at this point, so all IDs will be most likely 1

```
	if bg.GetRoutingStrategy() == MulticallRoutingStrategy && isValidMulticallTx(rpcReqs) && !isBatch {
		backendResp := bg.ExecuteMulticall(ctx, rpcReqs)
		return backendResp.RPCRes, backendResp.ServedBy, backendResp.error
	}

```

proxyd/backend.go
```diff
@@ -908,7 +911,31 @@ func (bg *BackendGroup) ProcessMulticallResponses(ch chan *multicallTuple, ctx c
...
+				if len(successResp) == 0 {
+					return errResp
+				}
+
+				// add rpc error responses to success responses
+				for id, rpcRes := range rpcResToCheck {
+					if _, ok := successResp[id]; !ok {
+						successResp[id] = rpcRes
+					}
+				}
```
jelias2: Whats the goal here? Is it to add errored responses to the successful responses map that match the same ID?

If all the requests have a similar ID of 1, it may not work as intended

jelias2 (commented):
Just curious on the desired outcome/behvaior of this feature, 

Given a request unique ID, does it intend to return the a success if it exists, otherwise it will return the error response.

 If there is an error and a success for the same ID, will it overwrite the error with the success?

proxyd/backend.go
jelias2: Thoughts on making   ProcessMulticallResponses and  ProcessMulticallResponsesWithErrors two different functions?

cody-wang-cb:
@jelias2 sorry I should've made the PR description more clear.

The motivation is that because we multicall `sendRawTransaction` to 2 different mempool nodes (geth + reth), we noticed sometimes one node would return an RPC error while the other return success with no error.

However at proxyd level, an RPC error is not an actual error since it's still `200`. This gives the end user a bad experience as we've gotten reports saying that they received an RPC error but the TX actually went through.

Let me know if the motivation makes sense

proxyd/backend.go
cody-wang-cb: Oh I see `isValidMulticallTx()` now.
I think that should work then if it's just 1 request each time, it would just only go through the loop once per check.

jelias2:
So with this new design, what do you wish to return to the user if there is one success and one failure for the same request?

If its all failures or all success, just return the appropriate success/failure


Overall no qualms about the implementation, I would lean placing the processErrorMulticall in a separate function and clearly document the what the behavior is for the same request returned with different results

cody-wang-cb:
> So with this new design, what do you wish to return to the user if there is one success and one failure for the same request?
> 
> If its all failures or all success, just return the appropriate success/failure
> 
> Overall no qualms about the implementation, I would lean placing the processErrorMulticall in a separate function and clearly document the what the behavior is for the same request returned with different results

if one success and one failure it'd return the success.
Sounds good let me update it as a separate function

proxyd/backend.go
```diff
@@ -931,6 +932,13 @@ func (bg *BackendGroup) ProcessMulticallResponses(ch chan *multicallTuple, ctx c
 			finalResp = resp
 			continue
 		}
+
+		// Assuming multicall doesn't support batch
+		if bg.multicallRPCErrorCheck && resp.RPCRes[0].IsError() {
```
cody-wang-cb, line 942: @jelias2 is this ok? Assuming multicall doesn't support batch

jelias2: Correct multi-call assumes single rpc request, no batches allowed

jelias2 (approved):

jelias2:
proxyd/v4.10.0 includes the new changes

PR #65, opened 2024-10-18
feat: add metric rpcReqTotal for every method

yjhmelody:


proxyd/metrics.go
```diff
@@ -504,6 +512,10 @@ func RecordCacheMiss(method string) {
 	cacheMissesTotal.WithLabelValues(method).Inc()
 }
 
+func RecordRpcRequestCount(method string, count int) {
+	rpcReqTotal.WithLabelValues(method).Add(float64(count))
+}
+
```
jelias2, line 518: Is it possible to use the Inc() function instead of Add(float64()) as it is more inline with the current proxyd style
```
rpcReqTotal.WithLabelValues(method).Inc()
```

jelias2:
Closing due to inactivity, please re-open if still requested with latest master

PR #71, opened 2024-10-25
feat(proxyd): forward optxproxy auth header

### LINKED ISSUE #68: [proxyd] optional to forward http headers with request
**Is your feature request related to a problem? Please describe.**
op-txproxy relies on http headers to validate authentication. proxy strips the headers when making the json-rpc request, leading to authentication errors on all requests

**Describe the solution you'd like**
Setup an option to forward request headers with the backend. Specified optionally since maybe it's a feature to strip headers in proxyd as well

hamdiallam:
**Description**
Closes #68.

Forwards the op-txproxy auth headers along with the request such that backends that rely on them work as expected

**Tests**

integration test

### DIFF
--- a/proxyd/backend.go
+++ b/proxyd/backend.go
@@ -561,6 +561,11 @@ func (b *Backend) doForward(ctx context.Context, rpcReqs []*RPCReq, isBatch bool
 		httpReq.SetBasicAuth(b.authUsername, b.authPassword)
 	}
 
+	opTxProxyAuth := GetOpTxProxyAuthHeader(ctx)
+	if opTxProxyAuth != "" {
+		httpReq.Header.Set(DefaultOpTxProxyAuthHeader, opTxProxyAuth)
+	}
+
 	xForwardedFor := GetXForwardedFor(ctx)
 	if b.stripTrailingXFF {
 		xForwardedFor = stripXFF(xForwardedFor)
--- a/proxyd/integration_tests/optxproxy_auth_test.go
+++ b/proxyd/integration_tests/optxproxy_auth_test.go
@@ -1,0 +1,55 @@
+package integration_tests
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"os"
+	"testing"
+
+	"github.com/ethereum-optimism/infra/proxyd"
+	"github.com/stretchr/testify/require"
+)
+
+func TestOpTxProxyAuthForwarder(t *testing.T) {
+	authEchoHandler := func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(200)
+		fmt.Fprintf(w, `{"jsonrpc": "2.0", "result": "%s", "id": 999}`, r.Header.Get(proxyd.DefaultOpTxProxyAuthHeader))
+	}
+
+	goodBackend := NewMockBackend(http.HandlerFunc(authEchoHandler))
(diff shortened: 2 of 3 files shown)


jelias2:
Is this supposed planning support a specifc type of authentication protocol or is supposed to be generalizable? 

Why does the authentication layer have to be at the proxyd / op-txproxy level as well? Typically proxyd is ran behind a simple ingress layer or reverse proxy service like traefik or nginx. I would think some of those layers have built in support for authentication protocols as well

Typically proxyd forwards to geth nodes or RPC providers with preconfigured auth token, so just curious where this request will be routed that it needs additional headers?

--------

Ahh I read the docs, and found the that only header needed is ```X-Optimism-Signature```. 
If thats the case, I think the code could be descoped to just focus on that specific header value

I would look towards storing the X-Optimsim-Signature value in the context for the request, and then passing that to the tx-proxy request, rather than forwarding all headers.


Some sections to look at
[populate-context](https://github.com/ethereum-optimism/infra/blob/c0aa2b8826ac9088810cb9f0d80e3bc4af2a7bf5/proxyd/server.go?plain=1#L619) to add the header to the context

[do-forward](https://github.com/ethereum-optimism/infra/blob/c0aa2b8826ac9088810cb9f0d80e3bc4af2a7bf5/proxyd/backend.go?plain=1#L564) adding the headers to the request

hamdiallam:
> Ahh I read the docs, and found the that only header needed is `X-Optimism-Signature`. If thats the case, I think the code could be descoped to just focus on that specific header value

Open to scoping down & being more specific. The reason I had it forward all header values was just thinking through the general "proxy" abstraction in networking. It seems like JSON-RPC doesn't rely on http headers so it's okay but I don't think proxies in general are suppose to removing any headers present on the request

https://www.w3.org/TR/ct-guidelines/#sec-altering-header-values

```
4.1.5 Alteration of HTTP Header Field Values
Other than the modifications required by [[RFC 2616 HTTP]](https://www.w3.org/TR/ct-guidelines/#ref-HTTP) proxies should not modify the values of header fields other than the User-Agent, Accept, Accept-Charset, Accept-Encoding, and Accept-Language header fields and must not delete header fields (see [4.1.5.5 Original Header Fields](https://www.w3.org/TR/ct-guidelines/#sec-original-headers)).
```

LMK what you think about this

jelias2:
I get the perspective on adding header support for general-purpose networking proxyd, however I think keeping the scope of proxyd focused on core ethereum rpc functionality is the preference at this point in time. I think if there continues to be demand for more header support then allowing all is acceptable, but for just one specific header. I'd rather just use the existing patterns in the code.

One question back at the the tx-proxy, would be why is the signature implemented in the header and not as a parameter in the rpc body?

hamdiallam:
> I get the perspective on adding header support for general-purpose networking proxyd, however I think keeping the scope of proxyd focused on core ethereum rpc functionality is the preference at this point in time. I think if there continues to be demand for more header support then allowing all is acceptable, but for just one specific header. I'd rather just use the existing patterns in the code.
> 
> One question back at the the tx-proxy, would be why is the signature implemented in the header and not as a parameter in the rpc body?

Cool sounds good to me.

The `eth_sendRawTransactionConditional` method interface is common across arb/poly/op hence not changing it. Also follows the same structure to how flashbots authenticates their endpoints via http headers

jelias2 (approved):

PR #90, opened 2024-11-08
feat: add fallback rate limiter and cache to fallback to memory if redis fails

meyer9:
**Description**

Adds a new config to proxyd: `redis.fallback_to_memory`.

This allows falling back to a memory implementation of Redis features if they fail instead of falling back to an error (in the case of rate limiting) or a no-op (in the case of caching).

I also moved redis connection testing which will warn if we're not able to connect with the new flag is enabled, but still error if the fallback is disabled.

If the context is cancelled for the cache, we don't retry with the secondary source because this indicates a 

**Tests**

Added test for rate limiter fallback and cache fallback.

### DIFF
--- a/proxyd/frontend_rate_limiter.go
+++ b/proxyd/frontend_rate_limiter.go
@@ -157,10 +157,6 @@ func NewFallbackRateLimiter(primary FrontendRateLimiter, secondary FrontendRateL
 
 func (r *FallbackRateLimiter) Take(ctx context.Context, key string) (bool, error) {
 	if ok, err := r.primary.Take(ctx, key); err != nil {
-		// if context is cancelled, return immediately
-		if err == context.Canceled {
-			return ok, err
-		}
 		return r.secondary.Take(ctx, key)
 	} else {
 		return ok, err
--- a/proxyd/frontend_rate_limiter.go
+++ b/proxyd/frontend_rate_limiter.go
@@ -54,7 +54,8 @@ func (l *limitedKeys) Take(key string, max int) bool {
 // truncated timestamp at which the struct was created. If
 // the current truncated timestamp doesn't match what's
 // referenced, the limit is reset. Otherwise, values in
-// a map are incremented to represent the limit.
+// a map are incremented to represent the limit. This will
+// never return an error.
 type MemoryFrontendRateLimiter struct {
 	currGeneration *limitedKeys
 	dur            time.Duration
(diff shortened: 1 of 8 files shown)


proxyd/frontend_rate_limiter.go
```diff
@@ -54,7 +54,8 @@ func (l *limitedKeys) Take(key string, max int) bool {
 // truncated timestamp at which the struct was created. If
 // the current truncated timestamp doesn't match what's
 // referenced, the limit is reset. Otherwise, values in
-// a map are incremented to represent the limit.
+// a map are incremented to represent the limit. This will
+// never return an error.
```
jelias2, line 58: Can the error variable return on line 73, 89 be removed then if it will forsure never return an error?
```
func (m *MemoryFrontendRateLimiter) Take(ctx context.Context, key string) (bool, error) {
```

proxyd/redis.go
```diff
@@ -13,10 +13,15 @@ func NewRedisClient(url string) (*redis.Client, error) {
...
 	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
 	defer cancel()
 	if err := client.Ping(ctx).Err(); err != nil {
-		return nil, wrapErr(err, "error connecting to redis")
+		return wrapErr(err, "error connecting to redis")
 	}
-	return client, nil
+
+	return nil
 }
```
jelias2, line 27: Question: Can proxyd parse specific error types from Redis responses to implement appropriate error handling strategies for each case?

I could see a case where if proxyd is unable to authenticate with Redis on initalization, proxyd may want to fail fast, instead of continuing with In-Memory caching.

Example does a bad password, or malformed ip address fail at the redis.NewClient level or Redis.Ping Level? Is it possible for proxyd handle those errors and fail fast?

proxyd/proxyd.go
```diff
@@ -67,6 +74,14 @@ func Start(config *Config) (*Server, func(), error) {
 		if err != nil {
 			return nil, nil, err
 		}
+
+		if err := CheckRedisConnection(redisClient); err != nil {
+			if config.Redis.FallbackToMemory {
+				log.Warn("failed to connect to redis, may fall back to in-memory cache: %v", "err", err)
```
jelias2, line 80: Check all the log.Warns. looks like the %v may be unnecessary

jelias2:
Just curious what specific error from redis instance when running proxyd that prompted this work?

The reason I ask because if the redis error is intermittent, I could see some narrow edge cases where the primary and fallback cache's could thrashed back and forth. Since state is not aligned between the two cache's this could produce unexpected behavior

jelias2:
Interesting work overall, thanks for opening the PR

proxyd/frontend_rate_limiter.go
meyer9: This matches the FrontendRateLimiter interface. Could move it to a separate private method `take` that never returns an error, but it's pretty easy to tell from this code that it never returns an error.

proxyd/redis.go
meyer9: I checked and it does still exit if the URL is completely invalid:

```
go run ./cmd/proxyd julian.toml
{"time":"2024-11-14T10:25:11.852071-08:00","level":"INFO","msg":"starting proxyd","version":"","commit":"","date":""}
{"time":"2024-11-14T10:25:11.85292-08:00","level":"ERROR+4","msg":"error starting proxyd","err":"redis: invalid URL scheme: "}
exit status 1
```

It's hard to distinguish between a weird looking URL and a network failure. When testing, I got "no such host", "i/o timeout" and "connection refused". Not sure if any of these are worth handling specifically here. We don't want to crash if the connection with redis is unavailable, and all of these could be caused by network issues rather than configuration issues.

proxyd/proxyd.go
meyer9: sorry, good catch! I think I originally had the %v and then realized it was the incorrect format so added "err", but forgot to remove the %v.

meyer9:
> Just curious what specific error from redis instance when running proxyd that prompted this work?
> 
> The reason I ask because if the redis error is intermittent, I could see some narrow edge cases where the primary and fallback cache's could thrashed back and forth. Since state is not aligned between the two cache's this could produce unexpected behavior

Good question!

We want to make sure if the redis cache fails or networking to the redis cache fails, requests don't start failing. The issue right now is that if the redis client fails, requests fail due to rate limiting (failing closed instead of failing open).

I agree the intermittent error issue could be a problem, but the failure case is actually better now than before - instead of refusing all requests due to rate limiting, the server will make an attempt at processing the requests (even if the cache is missed). Rate limiting will continue to apply on a per-instance level.

meyer9:
Moving to #92

PR #92, opened 2024-11-14
feat: add fallback rate limiter and cache to fallback to memory if redis fails

meyer9:
Original discussion here: https://github.com/ethereum-optimism/infra/pull/90

**Description**

Adds a new config to proxyd: `redis.fallback_to_memory`.

This allows falling back to a memory implementation of Redis features if they fail instead of falling back to an error (in the case of rate limiting) or a no-op (in the case of caching).

I also moved redis connection testing which will warn if we're not able to connect with the new flag is enabled, but still error if the fallback is disabled.

If the context is cancelled for the cache, we don't retry with the secondary source because this indicates a 

**Tests**

Added test for rate limiter fallback and cache fallback.

### DIFF
--- a/proxyd/cache.go
+++ b/proxyd/cache.go
@@ -55,10 +55,6 @@ func newFallbackCache(primaryCache Cache, secondaryCache Cache) *fallbackCache {
 
 func (c *fallbackCache) Get(ctx context.Context, key string) (string, error) {
 	val, err := c.primaryCache.Get(ctx, key)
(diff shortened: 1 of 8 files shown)


jelias2 (approved):
Lgtm

jelias2:
[proxyd/v4.11.0](https://github.com/ethereum-optimism/infra/releases/tag/proxyd%2Fv4.11.0)

PR #98, opened 2024-12-03
fix: exclude context canccel error from forward error counting

Revolution1:
**Description**

Proxyd bans backends based on its error rate.
But it even counts context canceled error as a forwarding error,
while the cancellation is actually triggered by incomplete HTTP requests from the client.

This means hackers can make DoS attacks by just sendinga bunch of requests and closing without waiting for a response.

**Tests**

We have an environment that client sends a lot this kind of requests to test connection or some thing.
**Before this patch:**
  the proxy will have no backends to serve in a very short time.
**After:**
  works like a charm.

BTW:
`wrapErr()` will strip the error's type.
It's recommended to use [this package](https://github.com/pkg/errors) to do error wrapping and assertion.

### DIFF
--- a/proxyd/backend.go
+++ b/proxyd/backend.go
@@ -1333,6 +1333,9 @@ type LimitedHTTPClient struct {
 
 func (c *LimitedHTTPClient) DoLimited(req *http.Request) (*http.Response, error) {
 	if err := c.sem.Acquire(req.Context(), 1); err != nil {
+		if errors.Is(err, context.Canceled) {
+			return nil, err
+		}
 		tooManyRequestErrorsTotal.WithLabelValues(c.backendName).Inc()
 		return nil, wrapErr(err, "too many requests")
 	}
--- a/proxyd/backend.go
+++ b/proxyd/backend.go
@@ -381,6 +381,10 @@ func (b *Backend) Forward(ctx context.Context, reqs []*RPCReq, isBatch bool) ([]
 			"method", metricLabelMethod,
 		)
 		res, err := b.doForward(ctx, reqs, isBatch)
+		// if it's canceld, we don't want to count it as an error
+		if strings.Contains(err.Error(), "context canceled") {
+			return nil, err
+		}
 		switch err {
 		case nil: // do nothing
 		case ErrBackendResponseTooLarge:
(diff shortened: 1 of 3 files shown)


atenjin (approved):

proxyd/backend.go
```diff
@@ -583,6 +585,10 @@ func (b *Backend) doForward(ctx context.Context, rpcReqs []*RPCReq, isBatch bool
 	start := time.Now()
 	httpRes, err := b.client.DoLimited(httpReq)
 	if err != nil {
+		// if it's canceld, we don't want to count it as an error
+		if errors.Is(err, context.Canceled) {
+			return nil, err
+		}
```
jelias2, line 591: Would it be possible to add a metric for canceled contexts here?

proxyd/backend_test.go
```diff
@@ -20,3 +26,34 @@ func TestStripXFF(t *testing.T) {
 		assert.Equal(t, test.out, actual)
 	}
 }
+
+func TestForwardContextCanceled(t *testing.T) {
```
jelias2, line 30: Looks good, would it be possible to add a few more tests and cases such as incrementing the intermittentErrorsSlidingWindow first and then checking that they are not incremented again. 

Maybe asset that the debug log exists too in the case of context canceled.

Also test the multicall routing strategy

proxyd/backend.go
```diff
@@ -583,6 +585,10 @@ func (b *Backend) doForward(ctx context.Context, rpcReqs []*RPCReq, isBatch bool
 	start := time.Now()
 	httpRes, err := b.client.DoLimited(httpReq)
 	if err != nil {
+		// if it's canceld, we don't want to count it as an error
```
jelias2, line 588: nit; canceld -> canceled

jelias2:
Left a few comments, overall looks good! Thanks for the contribution @Revolution1!

jelias2:
Stale PR

Revolution1:
@jelias2 sorry, forgot to check my notification for all this time. Is this still a bug in proxyd? If so, I'm ok to reopen the pr to catch up the latest master.

proxyd/backend.go
Revolution1: Can, but better in different pr. This is just a fix.

Issue #99, opened 2024-12-05
Add Redis cluster client support for Consensus mechanism


Vui-Chee:
In my organization, we use Redis cluster provider. But when we deploy current proxyd (x2) using a single-node Redis client, `redis.NewClient`, we run into issues like:

![Image](https://github.com/user-attachments/assets/61103e0b-4dc8-46a6-8efc-01631afde8a0)
(note: this is not the logs from the actual deployment, but reproduced locally)

So we made changes to the Redis client internally to use `redis.NewClusterClient` to support Redis cluster and the issue goes away.

Though I may have reservations about using Redis cluster for the HA consensus mechanism, **could you add support for Redis cluster**?

In my opinion, the consensus mechanism will be broken if the requests are routed inconsistently to different Redis instances, but if the routing strategy is to forward requests by the Redis key (certain key requests are sharded to the same instance), the consensus mechanism will not be affected.

I would also like to know your thoughts about this.

Vui-Chee:
I have a forked PR: https://github.com/okx/infra/pull/4, suggesting some of these changes. You could take a look and let me know your thoughts 🙏.

Vui-Chee:
Closed by #100.

PR #100, opened 2024-12-05
proxyd: add Redis cluster support

Vui-Chee:
**Description**

Main motivation is to able to connect to Redis cluster using `redis.NewClusterClient`  that will be use with consensus feature. But since go-redis supports `redis.UniversalClient`, we can also retype some params to allow this option in places like cache and rate limiting.

**Tests**

Ran a local Redis cluster (x3 instances) and ran local build of proxyd on test config. Used nginx to round robin between 2 local proxyd instances connected to the same Redis cluster to test the consensus mechanism (using local script).

**Metadata**
Added discussion in #99.

### DIFF
--- a/proxyd/config.go
+++ b/proxyd/config.go
@@ -36,8 +36,6 @@ type CacheConfig struct {
 	TTL     TOMLDuration `toml:"ttl"`
 }
 
-type RedisClientChoice string
-
 type RedisConfig struct {
 	// If `redis_cluster = true`, you can specify url string for multi-node cluster:
 	//    "redis://<user>:<password>@<host>:<port>?addr=<host2>:<port2>&addr=<host3>:<port3>"
--- a/proxyd/config.go
+++ b/proxyd/config.go
@@ -36,11 +36,24 @@ type CacheConfig struct {
 	TTL     TOMLDuration `toml:"ttl"`
 }
 
+type RedisClientChoice string
+
+const (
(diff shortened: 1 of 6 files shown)


proxyd/config.go
```diff
@@ -36,11 +36,24 @@ type CacheConfig struct {
...
+	// If choice = "cluster" is specified (toml), you can specify url string for multi-node cluster:
+	//    "redis://<user>:<password>@<host>:<port>?addr=<host2>:<port2>&addr=<host3>:<port3>"
+	// OR "rediss://<user>:<password>@<host>:<port>?addr=<host2>:<port2>&addr=<host3>:<port3>"
+	//
+	// Otherwise, it is also possible to specify single url for Redis cluster with proxy support.
+	URL              string            `toml:"url"`
+	Namespace        string            `toml:"namespace"`
+	ReadURL          string            `toml:"read_url"`
+	FallbackToMemory bool              `toml:"fallback_to_memory"`
+	Choice           RedisClientChoice `toml:"choice"`
```
jelias2: Could choice be updated to be a boolean since there are only two options, and can it be renamed to RedisCluster

jelias2 (commented):
Looks good, small nit on the naming! Thanks for the contribution @Vui-Chee!

proxyd/config.go
Vui-Chee: Done 👌!

jelias2 (approved):

PR #102, opened 2024-12-09
Make proxyd error handling more generic for different backend impleme…

angel-ding-cb:
**Description**

Currently proxyd's special error categorization is too specific to geth-based implementations and misses important error cases from other backend implementations (reth, nethermind, etc). This limits proxyd's ability to properly monitor and alert on important error conditions across different backend types.

**Solution**
Expand proxyd's error pattern matching to cover more error cases and variations in how different backends report them

### DIFF
--- a/proxyd/metrics.go
+++ b/proxyd/metrics.go
@@ -203,6 +203,17 @@ var (
 		"gas price too high",
 		"gas price too low",
 		"invalid parameters",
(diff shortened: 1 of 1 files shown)


jelias2 (approved):

jelias2:
proxyd/v4.12.0

PR #105, opened 2024-12-10
feat(proxyd): add filter polling support via RPC

ArtificialPB:
**Description**

Currently, the backend on which the filter was created at can be different than the backend being polled for changes (due to load balancing), resulting in `filter not found` errors. 

This PR adds support for creating filters and polling its changes via RPC calls when there are multiple backends defined. For each new filter, the server now stores which backend served the request. When filter changes are requested, the server correctly routes the request to the same backend where the filter was created at.

**Tests**

Added an integration test that verifies that the filter requests are being routed to the same backend.

**Additional context**

WS works fine because it keeps open the connection to the correct backend.

### DIFF
--- a/proxyd/backend.go
+++ b/proxyd/backend.go
@@ -763,12 +763,12 @@ func (bg *BackendGroup) Primaries() []*Backend {
 }
 
 // NOTE: BackendGroup Forward contains the log for balancing with consensus aware
-func (bg *BackendGroup) Forward(ctx context.Context, rpcReqs []*RPCReq, isBatch bool) ([]*RPCRes, string, error) {
+func (bg *BackendGroup) Forward(ctx context.Context, rpcReqs []*RPCReq, isBatch bool, targetBackend string) ([]*RPCRes, string, error) {
 	if len(rpcReqs) == 0 {
 		return nil, "", nil
 	}
 
-	backends := bg.orderedBackendsForRequest()
+	backends := bg.orderedBackendsForRequest(targetBackend)
 
 	overriddenResponses := make([]*indexedReqRes, 0)
 	rewrittenReqs := make([]*RPCReq, 0, len(rpcReqs))
@@ -999,13 +999,19 @@ func weightedShuffle(backends []*Backend) {
 	weightedshuffle.ShuffleInplace(backends, weight, nil)
 }
 
-func (bg *BackendGroup) orderedBackendsForRequest() []*Backend {
-	if bg.Consensus != nil {
+func (bg *BackendGroup) orderedBackendsForRequest(targetBackend string) []*Backend {
+	// ignore consensus load balancing if targetBackend is set - no backend other than targetBackend can
+	// serve the requests
+	if bg.Consensus != nil && targetBackend == "" {
 		return bg.loadBalancedConsensusGroup()
 	} else {
 		healthy := make([]*Backend, 0, len(bg.Backends))
 		unhealthy := make([]*Backend, 0, len(bg.Backends))
 		for _, be := range bg.Backends {
+			if targetBackend != "" && be.Name != targetBackend {
+				continue
+			}
+
 			if be.IsHealthy() {
 				healthy = append(healthy, be)
 			} else {
--- a/proxyd/config.go
+++ b/proxyd/config.go
@@ -29,6 +29,9 @@ type ServerConfig struct {
 	EnablePprof           bool `toml:"enable_pprof"`
 	EnableXServedByHeader bool `toml:"enable_served_by_header"`
 	AllowAllOrigins       bool `toml:"allow_all_origins"`
+
+	// FilterTimeoutSeconds specifies the maximum time to keep a filter for that has not been polled for changes.
+	FilterTimeoutSeconds int `toml:"filter_timeout_seconds"`
 }
 
 type CacheConfig struct {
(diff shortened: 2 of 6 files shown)


proxyd/server.go
```diff
@@ -548,6 +598,20 @@ func (s *Server) handleBatchRPC(ctx context.Context, reqs []json.RawMessage, isL
 			for i := range elems {
 				responses[elems[i].Index] = res[i]
 
+				req := elems[i].Req
+				if res[i].Result != nil && strings.Contains(req.Method, "eth_new") && strings.Contains(req.Method, "Filter") {
+					f := &filter{
+						backendGroup: group.backendGroup,
+						backendName:  strings.SplitN(sb, "/", 2)[1],
```
ArtificialPB, line 605: `sb` has the following format: `fmt.Sprintf("%s/%s", bg.Name, back.Name)` (see `proxyd/backend.go:1410`)

proxyd/backend.go
```diff
@@ -763,12 +763,12 @@ func (bg *BackendGroup) Primaries() []*Backend {
...
 
 // NOTE: BackendGroup Forward contains the log for balancing with consensus aware
-func (bg *BackendGroup) Forward(ctx context.Context, rpcReqs []*RPCReq, isBatch bool) ([]*RPCRes, string, error) {
+func (bg *BackendGroup) Forward(ctx context.Context, rpcReqs []*RPCReq, isBatch bool, targetBackend string) ([]*RPCRes, string, error) {
 	if len(rpcReqs) == 0 {
 		return nil, "", nil
 	}
 
-	backends := bg.orderedBackendsForRequest()
+	backends := bg.orderedBackendsForRequest(targetBackend)
```
jelias2, line 771: It it even worth it to pass the name into bg.orderedBackendsForRequest? Could it just look up the target backend name and skip orderedBackendForRequest

proxyd/config.go
```diff
@@ -29,6 +29,9 @@ type ServerConfig struct {
 	EnablePprof           bool `toml:"enable_pprof"`
 	EnableXServedByHeader bool `toml:"enable_served_by_header"`
 	AllowAllOrigins       bool `toml:"allow_all_origins"`
+
+	// FilterTimeoutSeconds specifies the maximum time to keep a filter for that has not been polled for changes.
+	FilterTimeoutSeconds int `toml:"filter_timeout_seconds"`
```
jelias2, line 34: Please make this feature of filter tracking configurable on and off with a boolean

proxyd/server.go
```diff
@@ -486,11 +502,45 @@ func (s *Server) handleBatchRPC(ctx context.Context, reqs []json.RawMessage, isL
...
+
+				group = f.backendGroup
+				backendName = f.backendName
+			} else {
+				RecordRPCError(ctx, BackendProxyd, parsedReq.Method, err)
+				responses[i] = NewRPCErrorRes(parsedReq.ID, ErrInvalidParams("filter not found"))
+				continue
+			}
+		}
+
```
jelias2, line 538: Please encapsulate this log into a function

proxyd/server.go
```diff
@@ -654,6 +718,27 @@ func randStr(l int) string {
...
+	ticker := time.NewTicker(s.filterTimeout)
+	defer ticker.Stop()
+	for {
+		<-ticker.C
+		s.filtersMu.Lock()
+		for id, f := range s.filters {
+			select {
+			case <-f.deadline.C:
+				// just delete, the node will automatically expire it on their end
+				delete(s.filters, id)
```
jelias2, line 733: What is the default filter timeout on the node itself? 
Is the user of this RPC able to override this? 
Adding too many filters to a node may slow down performance and not actually deleting in the filter timeout by proxyd could be a concern

proxyd/server.go
```diff
@@ -548,6 +598,20 @@ func (s *Server) handleBatchRPC(ctx context.Context, reqs []json.RawMessage, isL
 			for i := range elems {
 				responses[elems[i].Index] = res[i]
 
+				req := elems[i].Req
```
jelias2, line 601: Please put 601/614 in a function

jelias2 (commented):
Thanks for the great PR and contribution. I've left some initial comments, and I'd also like another person from out team to review this. 

Currently we really dont use proxyd with filters so I'm hesitant to merge and this functionality as our team will need to maintain this.

proxyd/backend.go
```diff
@@ -763,12 +763,12 @@ func (bg *BackendGroup) Primaries() []*Backend {
 }
 
 // NOTE: BackendGroup Forward contains the log for balancing with consensus aware
-func (bg *BackendGroup) Forward(ctx context.Context, rpcReqs []*RPCReq, isBatch bool) ([]*RPCRes, string, error) {
+func (bg *BackendGroup) Forward(ctx context.Context, rpcReqs []*RPCReq, isBatch bool, targetBackend string) ([]*RPCRes, string, error) {
```
jelias2, line 766: It its possible to add targeted backend to the context of the request instead of passing in another argument?

ArtificialPB:
@jelias2 thank you for the context around supporting filters.

I'll wait for additional feedback, if this is something you want to support in proxyd, before addressing the points from your review.

jelias2:
Hi Kriptal,
The proposal did not gain much traction internally, so at this time I don't think its appropriate to merge.

Thanks for the contribution and in the future if you want to voice your thoughts or feedback, feel free to join the OP Protocol R&D Discord. I'm frequently in #platforms if you want quicker feedback 

[Invite Link](https://discord.gg/B4QkMvAP)

PR #106, opened 2024-12-10
feat: add short-lived cache handlers

meyer9:
**Description**

We currently send a ton of traffic to backend nodes on barely uncacheable endpoints (endpoints based on block number). This PR adds a cache with a very short TTL (`block_ttl` - defaults to `2 seconds`) for these endpoints.

> Since Redis 2.6 the expire error is from 0 to 1 milliseconds.

Short TTLs seem fine for Redis. Other caching backends will currently ignore short lived cache keys.

_Open question: should we implement this for in-memory cache?_

**Tests**

TODO

**Additional context**

### DIFF
--- a/proxyd/cache.go
+++ b/proxyd/cache.go
@@ -3,6 +3,7 @@ package proxyd
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"strings"
 	"time"
 
@@ -100,6 +101,7 @@ func (c *redisCache) Get(ctx context.Context, key string) (string, error) {
 	start := time.Now()
 	val, err := c.redisReadClient.Get(ctx, c.namespaced(key)).Result()
 	redisCacheDurationSumm.WithLabelValues("GET").Observe(float64(time.Since(start).Milliseconds()))
+	fmt.Println("val", val)
 
 	if err == redis.Nil {
 		return "", nil
@@ -114,8 +116,11 @@ func (c *redisCache) Put(ctx context.Context, key string, value string, shortLiv
 	ttl := c.defaultTTL
 
 	// disable PUT on short lived key if shortLivedTTL is not set
-	if shortLived && c.shortLivedTTL == 0 {
-		return nil
+	if shortLived {
+		if c.shortLivedTTL == 0 {
+			return nil
+		}
+		ttl = c.shortLivedTTL
 	}
 
 	start := time.Now()
--- a/proxyd/cache.go
+++ b/proxyd/cache.go
@@ -110,8 +110,13 @@ func (c *redisCache) Get(ctx context.Context, key string) (string, error) {
 }
 
 func (c *redisCache) Put(ctx context.Context, key string, value string, shortLived bool) error {
+	ttl := c.defaultTTL
+	if shortLived {
+		ttl = c.shortLivedTTL
+	}
+
 	start := time.Now()
-	err := c.redisClient.SetEx(ctx, c.namespaced(key), value, c.defaultTTL).Err()
+	err := c.redisClient.SetEx(ctx, c.namespaced(key), value, ttl).Err()
 	redisCacheDurationSumm.WithLabelValues("SETEX").Observe(float64(time.Since(start).Milliseconds()))
 
 	if err != nil {
--- a/proxyd/cache.go
+++ b/proxyd/cache.go
@@ -3,7 +3,6 @@ package proxyd
 import (
 	"context"
 	"encoding/json"
-	"fmt"
 	"strings"
 	"time"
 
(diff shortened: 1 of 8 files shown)


proxyd/cache.go
```diff
@@ -101,9 +109,9 @@ func (c *redisCache) Get(ctx context.Context, key string) (string, error) {
 	return val, nil
 }
 
-func (c *redisCache) Put(ctx context.Context, key string, value string) error {
+func (c *redisCache) Put(ctx context.Context, key string, value string, shortLived bool) error {
 	start := time.Now()
-	err := c.redisClient.SetEx(ctx, c.namespaced(key), value, c.ttl).Err()
+	err := c.redisClient.SetEx(ctx, c.namespaced(key), value, c.defaultTTL).Err()
```
mdehoog: this should pass `shortLivedTTL` if `shortLived` is `true`?

meyer9: ah yeah, fixed that!

proxyd/config.go
```diff
@@ -32,8 +32,9 @@ type ServerConfig struct {
 }
 
 type CacheConfig struct {
-	Enabled bool         `toml:"enabled"`
-	TTL     TOMLDuration `toml:"ttl"`
+	Enabled  bool         `toml:"enabled"`
+	TTL      TOMLDuration `toml:"ttl"`
+	BlockTTL TOMLDuration `toml:"block_ttl"`
```
jelias2: block_ttl could probably be changed to something live ShortLivedTTL to better reflect it can be used for multiple RPCs

proxyd/cache.go
```diff
@@ -176,14 +194,24 @@ func newRPCCache(cache Cache) RPCCache {
...
+		"eth_getBlockByNumber":                    shortLivedHandler,
+		"eth_blockNumber":                         shortLivedHandler,
+		"eth_getBalance":                          shortLivedHandler,
+		"eth_getStorageAt":                        shortLivedHandler,
+		"eth_getTransactionCount":                 shortLivedHandler,
+		"eth_getBlockTransactionCountByNumber":    shortLivedHandler,
+		"eth_getUncleCountByBlockNumber":          shortLivedHandler,
+		"eth_getCode":                             shortLivedHandler,
+		"eth_getTransactionByBlockNumberAndIndex": shortLivedHandler,
+		"eth_getUncleByBlockNumberAndIndex":       shortLivedHandler,
```
jelias2, line 215: Should all of these be low TTL? ex. eth_getCode, eth_getBlockByNumber or eth_getBlockTransactionCountByNumber seem like they would return consistent results ? 

Just curious how these RPCs were chosen, any related data? I get that tip relevant RPCs and rapidly changing data fit the low TTL behavior

jelias2: Follow up could be making this configurable but I think that maybe out of scope for the inital PR

proxyd/proxyd.go
```diff
@@ -310,7 +310,12 @@ func Start(config *Config) (*Server, func(), error) {
 			if config.Cache.TTL != 0 {
 				ttl = time.Duration(config.Cache.TTL)
 			}
-			cache = newRedisCache(redisClient, redisReadClient, config.Redis.Namespace, ttl)
+
+			blockTtl := defaultBlockTtl
+			if config.Cache.BlockTTL != 0 {
+				blockTtl = time.Duration(config.Cache.BlockTTL)
+			}
+			cache = newRedisCache(redisClient, redisReadClient, config.Redis.Namespace, ttl, blockTtl)
```
jelias2: Would it be possible to write the feature so the shortLivedTTL can be toggled on and off? With the default being off? It looks like in its current state shortLivedTTL will always be enabled

jelias2 (commented):
Overall looks good, definitely can get this merged! Left a couple comments. PTAL, and in the final form please remove the prints, and add a couple more UTs if the feature can be toggled on and off.

Thanks for the contribution @meyer9!

jelias2:
Currently the PR is still in draft, please update once its ready. 
Also could the readme be updated under cacheable method with a section about the shortlived TTL

proxyd/cache.go
```diff
@@ -101,7 +101,7 @@ func (c *redisCache) Get(ctx context.Context, key string) (string, error) {
 	start := time.Now()
 	val, err := c.redisReadClient.Get(ctx, c.namespaced(key)).Result()
 	redisCacheDurationSumm.WithLabelValues("GET").Observe(float64(time.Since(start).Milliseconds()))
-	fmt.Printf("GET key: %s, val: %s\n", key, val)
+	fmt.Println("val", val)
```
jelias2: nit, please remove

PR #114, opened 2024-12-25
[proxyd]: Add block range rate limiting, and max block range to non-consensus mode

mdehoog:
**Description**

This PR implements two things:
 - the ability to limit the max block range, even if not in consensus mode
 - (opt-in, off by default) the ability to rate limit block range requests by block count

**Tests**

Added new tests.

**Additional context**

https://github.com/ethereum-optimism/optimism/pull/6686 originally implemented the ability to limit block range requests. The initial implementation supported both consensus and non-consensus mode, however there was feedback on that PR that we already had latest block polling in consensus mode, and any block range limiting should be added to that.

We've since tried to use consensus mode, and even opened some PRs to help (e.g. https://github.com/ethereum-optimism/optimism/pull/7275), but unfortunately our setup is not conducive to consensus mode.

### DIFF
--- a/proxyd/backend.go
+++ b/proxyd/backend.go
@@ -380,7 +380,7 @@ func (b *Backend) Forward(ctx context.Context, reqs []*RPCReq, isBatch bool) ([]
 			"max_attempts", b.maxRetries+1,
 			"method", metricLabelMethod,
 		)
-		res, err := b.doForward(ctx, reqs, isBatch)
+		res, err := b.doForward(ctx, reqs, isBatch, false)
 		switch err {
 		case nil: // do nothing
 		case ErrBackendResponseTooLarge:
@@ -454,6 +454,10 @@ func (b *Backend) ProxyWS(clientConn *websocket.Conn, methodWhitelist *StringSet
 
 // ForwardRPC makes a call directly to a backend and populate the response into `res`
 func (b *Backend) ForwardRPC(ctx context.Context, res *RPCRes, id string, method string, params ...any) error {
+	return b.forwardRPC(ctx, false, res, id, method, params...)
+}
+
+func (b *Backend) forwardRPC(ctx context.Context, isNonConsensusPoll bool, res *RPCRes, id string, method string, params ...any) error {
 	jsonParams, err := json.Marshal(params)
 	if err != nil {
 		return err
@@ -466,7 +470,7 @@ func (b *Backend) ForwardRPC(ctx context.Context, res *RPCRes, id string, method
 		ID:      []byte(id),
 	}
 
-	slicedRes, err := b.doForward(ctx, []*RPCReq{&rpcReq}, false)
+	slicedRes, err := b.doForward(ctx, []*RPCReq{&rpcReq}, false, isNonConsensusPoll)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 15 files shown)


jelias2:
Lmk if this is ready for review, typically would like to see CI passing as a prerequisite

proxyd/backend.go
```diff
@@ -482,9 +486,19 @@ func (b *Backend) ForwardRPC(ctx context.Context, res *RPCRes, id string, method
 	return nil
 }
 
-func (b *Backend) doForward(ctx context.Context, rpcReqs []*RPCReq, isBatch bool) ([]*RPCRes, error) {
+func (b *Backend) doForward(ctx context.Context, rpcReqs []*RPCReq, isBatch, isNonConsensusPoll bool) ([]*RPCRes, error) {
 	// we are concerned about network error rates, so we record 1 request independently of how many are in the batch
-	b.networkRequestsSlidingWindow.Incr()
+	// (we don't count non-consensus polling towards error rates)
```
mdehoog, line 491: The alternative to this is to disable the non-consensus poller in more tests by setting the `nonconsensus_poller_interval` negative... what do you think? To me it feels like we shouldn't be counting polling towards error rates, but I don't feel strongly.

mdehoog:
@jelias2 thank you 🙏 

Apologies, I thought tests were passing when I marked this as ready for review, but I now realize it was only the `check-updated-files` check. All the tests have been fixed and CI is now passing.

proxyd/block_range.go
```diff
@@ -0,0 +1,68 @@
...
+
+type BlockNumberTracker interface {
+	GetLatestBlockNumber() (uint64, bool)
+	GetSafeBlockNumber() (uint64, bool)
+	GetFinalizedBlockNumber() (uint64, bool)
+}
+
+func ExtractBlockRange(req *RPCReq, tracker BlockNumberTracker) *BlockRange {
+	switch req.Method {
+	case "eth_getLogs", "eth_newFilter":
```
jelias2, line 22: eth_newFilter is only appended to a single backend, so it quite difficult to get the results from the requesting the filter through proxyd. We typically don't expose it on proxyd on our end. Curious your experience with it

A while ago there was a PR for similar functionality: https://github.com/ethereum-optimism/infra/pull/105, but I didn't see any interest in at the time.

proxyd/config.go
```diff
@@ -182,6 +188,8 @@ type BackendGroupConfig struct {
 	ConsensusHALockPeriod        TOMLDuration `toml:"consensus_ha_lock_period"`
 	ConsensusHARedis             RedisConfig  `toml:"consensus_ha_redis"`
 
+	NonconsensusPollerInterval TOMLDuration `toml:"nonconsensus_poller_interval"`
```
jelias2, line 191: Overall I'm curious why the non-consensus poller is needed? If a backend was unable to serve a eth_getLogs in fallback mode, it would just request the next backend in the backend group? Whats the benefit of adding this component?

proxyd/frontend_rate_limiter.go
```diff
@@ -122,14 +118,14 @@ func (r *RedisFrontendRateLimiter) Take(ctx context.Context, key string) (bool,
 		return false, err
 	}
 
-	return incr.Val()-1 < int64(r.max), nil
+	return incr.Val() <= int64(r.max), nil
```
jelias2, line 121: Just curious why removing the -1

proxyd/nonconsensus_poller.go
```diff
@@ -0,0 +1,123 @@
...
+	value, err := p.fetchBlock(p.ctx, be, label)
+	if err != nil {
+		log.Error("failed to fetch block", "backend", be.Name, "label", label, "err", err)
+		return
+	}
+
+	p.lock.Lock()
+	defer p.lock.Unlock()
+
+	if *ptr < value || *ptr == math.MaxUint64 {
```
jelias2, line 106: Whats the purpose of *ptr == math.MaxUnit64 in the conditional? Is this if the ptr is uninitialzed?

proxyd/backend.go
jelias2: I think the sliding network requests sliding window should be counted. IRRC, its counted for the consensus_poller, I think it would be odd if it was counted for one strategy but not the other

jelias2: Additionally if the poll fails that helps proxyd determine if the backend is unstable at the moment, which is why I'd lean toward tracking the intermittent sliding window error as well

proxyd/block_range.go
jelias2: i guess eth_newFilter is used for overwriting the reqeust parameters, so  it makes sense to keep it in here, but still I don't think that method works perfectly

proxyd/config.go
jelias2: The consensus poller allows re-writing of latest, safe, and finalized, I guess the major drawback would be if a eth_getLogs request goes a backend which doesn't have the non-consensus latest if would be missing the latest logs. Could you provide a bit more background on the behavior that this PR aims to fix and the expected results?

jelias2:
Overall the blockrange rate limiting looks good if you want to open a second PR for that. For the max block range on non-consensus mode would like a bit more context. 

Thanks for the contribution!

PR #115, opened 2025-01-07
cci-stats: Add project

mslipper:


cci-stats/cmd/runner/main.go
```diff
@@ -0,0 +1,98 @@
...
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"os/signal"
+	"regexp"
+	"strconv"
+
+	"github.com/axelKingsley/go-circleci"
```
scharissis, line 11: These changes are tiny; should we try get them merged upstream? It's easy to forget dependencies like this and then they stagnate.

cci-stats/cmd/runner/main.go
```diff
@@ -0,0 +1,98 @@
...
+			cancel()
+		case err := <-errC:
+			return err
+		case <-ctx.Done():
+			return nil
+		}
+	}
+}
+
+func requiredStr(envVar string) string {
```
scharissis: nit
```suggestion
func requiredStrEnv(envVar string) string {
```

cci-stats/cmd/runner/main.go
```diff
@@ -0,0 +1,98 @@
...
+	val := os.Getenv(envVar)
+	if val == "" {
+		panic(fmt.Errorf("%s must be set", envVar))
+	}
+	return val
+}
+
+func requiredInt(envVar string) int {
```
scharissis: nit
```suggestion
func requiredIntEnv(envVar string) int {
```

cci-stats/cmd/runner/main.go
```diff
@@ -0,0 +1,98 @@
+package main
```
scharissis, line 1: It would be great to have a short comment, or readme, describing what `cci-stats` is/does.

cci-stats/cmd/runner/main.go
```diff
@@ -0,0 +1,98 @@
...
+			return nil
+		}
+	}
+}
+
+func requiredStr(envVar string) string {
+	val := os.Getenv(envVar)
+	if val == "" {
+		panic(fmt.Errorf("%s must be set", envVar))
+	}
```
scharissis, line 84: You may prefer to use `os.LookupEnv` to be able to identify the case where it is actually unset as opposed to an empty string.
https://pkg.go.dev/os#LookupEnv

(same goes for `requiredInt()` func)

cci-stats/pkg/db/db.go
```diff
@@ -0,0 +1,211 @@
+package db
```
scharissis, line 1: I'm not sure how important this service is, but there's a lot of untested code in here and I feel the need to call it out.

cci-stats/pkg/db/db.sql
```diff
@@ -0,0 +1,51 @@
+CREATE TABLE pipelines
```
scharissis, line 1: Is the file `db.sql` used anywhere here? Did you perhaps forget to include the migrator code?
If unused, should we keep it? It'll go stale.

cci-stats/pkg/service/service.go
```diff
@@ -0,0 +1,281 @@
...
+	}
+
+	if err := tx.Commit(ctx); err != nil {
+		return fmt.Errorf("failed to commit transaction: %w", err)
+	}
+
+	return nil
+}
+
+func fetchPipelines(ctx context.Context, config config.Config, cutoff time.Time, client *circleci.Client) ([]*circleci.Pipeline, error) {
```
scharissis, line 191: nit/idea: this could potentially be used to fetch the `LastPipeline`, thus simplifying reducing the db interface

scharissis: Also, another idea, this `cutoff` feels like it could perhaps be superseded by a pagination token.

cci-stats/cmd/runner/main.go
mslipper: The upstream go-circleci hasn't been maintained in 2 years. Axel works here, so I'm more confident in his fork than I am of the upstream.

cci-stats/cmd/runner/main.go
mslipper: I need a value to be specified for these env vars, so both being unset and set to an empty string are errors.

cci-stats/pkg/db/db.sql
mslipper: It's used to bootstrap the database. The database backing the stats on go/ci-dashboard was generated using this file. There is no migrator since the application is very simple.

cci-stats/pkg/db/db.go
mslipper: All this service does is populate the stats on http://go/ci-dashboard. Bugs in this code just make the stats show up wrong, and the entire project is just a hair over 500 SLOC so I didn't think it was worth investing much time in detailed unit testing particularly since the majority of the logic reads from the CCI API and would require mocking.

cci-stats/pkg/service/service.go
mslipper: `LastPipeline` is used to determine what the latest pipeline in the database is. The `createdAt` value from that pipeline is then used as the `cutoff` value. When the build with the `createdAt` value matching the `cutoff` value is found, the application stops collecting pipelines and starts collecting other data (e.g. tests, workflows, etc.). Since the local value matters here, we'll still need the DB interface to expose a getter for its latest pipeline.

How would I use the pagination token here? There's no way to know how many pages we need to read in advance, and the token itself is returned by the CCI API when you make a request for a pipeline.

scharissis (approved):

cci-stats/pkg/service/service.go
scharissis: I see, thanks. I was hoping the CCI API would expose a pagination token we could leverage. But I didn't grasp the particulars of the logic around `cutoff`. Was just spit-balling ideas. :)

scharissis (approved):

PR #131, opened 2025-01-28
feat(nat): dockerize nat.

scharissis:
**Description**

Containerized NAT.

**Tests**
Tested with our local docker-compose.
The tests pictured below currently fail as it can't access my local devnet (there appears to be something wrong with kurtosis-devnet at the moment).
But at least it shows it runs within a container.

Screenshot 2025-01-28 at 17 19 01

### DIFF
--- a/op-nat/Dockerfile
+++ b/op-nat/Dockerfile
@@ -1,0 +1,19 @@
+FROM golang:1.23.5-alpine3.21 AS builder
+
+WORKDIR /app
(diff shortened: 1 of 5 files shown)


jelias2:
LGTM, maybe just add README link to where the kurtosis devnet can be found in the monorepo, so that if an external person wants to run it they can

jelias2 (approved):

PR #161, opened 2025-02-04
op-conductor-ops: removing sequencers not in cluster

zhwrd:
* report sequencers in cluster but missing from config
* allow removing them using `remove-server`

### DIFF
--- a/op-conductor-ops/op-conductor-ops.py
+++ b/op-conductor-ops/op-conductor-ops.py
@@ -97,10 +97,14 @@ def status(network: str):
                     print_error(
                         f": {sequencer.sequencer_id} does not have the correct voting status.")
                     display_correction = True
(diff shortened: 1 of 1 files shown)


yashvardhan-kukreja (approved):

op-conductor-ops/op-conductor-ops.py
```diff
@@ -97,10 +97,14 @@ def status(network: str):
                     print_error(
                         f": {sequencer.sequencer_id} does not have the correct voting status.")
                     display_correction = True
+                del membership[sequencer.sequencer_id]
```
yashvardhan-kukreja, line 100: Slightly ambiguous naming of the dict imo. Something like `untracked_members` would be more relevant imo.

But definitely not worth blocking this PR. So feel free to address or ignore this comment :)

PR #166, opened 2025-02-06
feat(nat): run on a schedule.

### LINKED ISSUE #136: [nat] Run on a schedule
* Ensure it's running periodically
* Tune run period
* Ensure metrics are being emitted for each run

scharissis:
**Description**

Makes NAT run the tests on an hourly schedule.

**Metadata**

- Resolves #136

### DIFF
--- a/op-nat/nat.go
+++ b/op-nat/nat.go
@@ -5,12 +5,17 @@ import (
 	"errors"
 	"os"
 	"sync/atomic"
+	"time"
 
 	"github.com/google/uuid"
 	"github.com/jedib0t/go-pretty/v6/table"
 
 	"github.com/ethereum-optimism/infra/op-nat/metrics"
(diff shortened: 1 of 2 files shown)


op-nat/nat.go
```diff
@@ -13,6 +14,10 @@ import (
 	"github.com/ethereum-optimism/optimism/op-service/cliapp"
 )
 
+const (
+	runInterval = 1 * time.Hour
+)
+
```
jelias2, line 20: LGTM couple suggestions,

A flag or env var to configure this would be a nice touch. 
Also log the run the interval so the user don't have to look at the code to figure out the the interval

jelias2 (approved):

scharissis:
Closing; was done in another PR.

PR #171, opened 2025-02-13
[NAT] Pivot to being a test runner

### LINKED ISSUE #172: [nat] pivot to being a test runner
* Make nat discover and run tests from a directory
* Remove old code and tests

scharissis:
**Description**
This PR implements the pivot for NAT which makes it more focused on being a test runner.

Key features:
* Old acceptance tests removed
* Old code for tests/suites/gates removed
* Discovers tests (recursively) within a root directory
 * Tests can be an entire package
 * If a test doesn't have a package associated with it, NAT searches for it
* Handles test skips
* Nicer output table (handles skips, shows test duration, shows hierarchy)
* A "validator config" YAML is used to describe the relationship between tests/suites/gates. It's intended for this file to also live in the optimism monorepo (location TBD)
* It has a simple 'registry' which keeps tracks of the validator config; allows for programmatic/dynamic adjustments as well as validation
* Allows for gate inheritance (eg. Gate B can inherit all of Gate A's contents and simply add/edit them)

Video summary:
https://www.loom.com/share/e64a3236ea874fbdafcac67d8233249b?sid=249b72ef-d9eb-4ad1-9e9f-60cb4d82cc0c

CLI interface:
```bash
go run cmd/main.go \
  --testdir ../../optimism/ \
  --gate betanet \
  --manifest alpaca.json \
  --validators example-validators.yaml
```

Example Validators YAML:
```yaml
gates:
  - id: localnet
    description: "Localnet validation gate"
    suites:
      smoke:
        tests:
          - name: TestInteropSystemNoop
            package: github.com/ethereum-optimism/optimism/kurtosis-devnet/tests/interop
    tests:
      - name: TestFindRPCEndpoints
        package: github.com/ethereum-optimism/optimism/kurtosis-devnet/pkg/kurtosis/api/run


  - id: alphanet
    description: "Alphanet validation gate"
    inherits: ["localnet"]
    suites:
      devnet_system:
        description: "System tests"
        tests:
          - name: TestWallet
            package: github.com/ethereum-optimism/optimism/devnet-sdk/system
          - name: TestChainUser
            package: github.com/ethereum-optimism/optimism/devnet-sdk/system
      kurtosis:
        description: "Kurtosis tests"
        tests:
          - package: github.com/ethereum-optimism/optimism/kurtosis-devnet/pkg/kurtosis

  - id: betanet
    description: "Betanet validation gate"
    inherits: ["alphanet"]
    suites:
      interop:
        description: "Basic network functionality tests"
        tests:
          - name: TestInteropSystemNoop
            package: github.com/ethereum-optimism/optimism/kurtosis-devnet/tests/interop
          - name: FuzzDetectNonBijectivity
            package: github.com/ethereum-optimism/optimism/kurtosis-devnet/tests/interop
```

**Metadata**
Resolves #172


op-nat/README.md
```diff
@@ -1,54 +1,69 @@
...
+For an example `validators.yaml`, see [example-validators.yaml](./example-validators.yaml).
+
+**Gate**
+A gate is a collection of suites and/or tests that represent a complete validation scenario. A gate can inherit from one or more other gates.
+
+```yaml
+# Example gate definition
+- id: alphanet
+  description: "Alphanet validation gate"
+  inherits: ["localnet"]
```
jelias2: Does the inherits config have to be defined in the same yaml file at this point in time?

op-nat/README.md
```diff
@@ -1,54 +1,69 @@
...
-* Avoid using "test/suite/gate" in names as it's already implied by the package
+**Test**
+A test is a single check that validates a specific aspect of your network.
+A description is optional.
+A package is optional. If not provided, the test will be searched for. This lengthens the runtime, so we recommend providing the package.
+If a package is provided and no name is provided, NAT runs all tests in the package.
+
+```yaml
+# Run test 'TestInteropSystem' in package 'github.com/ethereum-optimism/optimism/kurtosis-devnet/tests/interop'
+- name: TestInteropSystem
```
jelias2: What about a config example for running a single test file within a package? Is that possible at this time?

op-nat/config.go
```diff
@@ -19,10 +21,13 @@ import (
...
-	SC         SuperchainManifest
-	Validators []Validator
+	SC SuperchainManifest
 
-	Wallets []*wallet.Wallet
+	// Test config
+	TestDir         string
+	ValidatorConfig string
+	TargetGate      string
+	Wallets         []*wallet.Wallet
```
jelias2: Are wallets still necessary in the new version? Shouldn't these be define in the test

op-nat/config.go
```diff
@@ -31,14 +36,34 @@ type Config struct {
...
 	if err := flags.CheckRequired(ctx); err != nil {
 		return nil, fmt.Errorf("missing required flags: %w", err)
 	}
+	if testDir == "" {
+		return nil, errors.New("test directory is required")
+	}
+	if validatorConfig == "" {
+		return nil, errors.New("validator config path is required")
+	}
+	if gate == "" {
```
jelias2: Does gate need to be required? What if the whole config file should be ran?

op-nat/config.go
scharissis: You're right. I'll try remove these as part of this PR/change. Thanks.

op-nat/README.md
scharissis: File-based isn't possible at this time. It's also unclear to me whether we want it.
For now we can express:
* Single test
* Entire package

op-nat/README.md
scharissis: Great question. Yes. For now, we only support a single 'validation config' file.

op-nat/config.go
scharissis: It's a good question. I figured it'd be better if we have to specify a gate.
Let me know if you want to chat about it though!

op-nat/config.go
scharissis: Done. Removed in 29434e5.

op-nat/README.md
jelias2: I think down the line, expressing an specific file would be a nice to have. Lets add it to the nat backlog

op-nat/config.go
```diff
@@ -1,55 +1,65 @@
...
 	// Network config
-	SC         SuperchainManifest
-	Validators []Validator
-
-	Wallets []*wallet.Wallet
-
-	// Networks
+	SC  SuperchainManifest
 	L1  *network.Network
 	L2A *network.Network
```
jelias2: I also believe. network objects will be defined in the tests so these can be removed as well

op-nat/config.go
jelias2: Sure feel free to leave a gate for now, however it would be nice to make nat comptabile with just running a single test. Thinking about a case where nat is running on a network and debugging a single failure,  It would be conveint for the developer to run a single test without the overhead of having to define a whole gate. 

Can be backlogged

op-nat/example-validators.yaml
```diff
@@ -0,0 +1,43 @@
...
+    description: "Alphanet validation gate"
+    inherits: ["localnet"]
+    suites:
+      devnet_system:
+        description: "System tests"
+        tests:
+          - name: TestWallet
+            package: github.com/ethereum-optimism/optimism/devnet-sdk/system
+          - name: TestChainUser
+            package: github.com/ethereum-optimism/optimism/devnet-sdk/system
```
jelias2: Just curious, are suites able to be defined outside of gate? 

Ex. to compose gates of multiple suites, without needing to inherit an sub gate?

op-nat/metrics/metrics.go
```diff
@@ -40,15 +42,51 @@ var (
 		"result",
 	})
 
-	acceptancesTotal = promauto.NewCounterVec(prometheus.CounterOpts{
+	acceptanceResults = promauto.NewGaugeVec(prometheus.GaugeOpts{
```
jelias2, line 45: For all of these metrics whats the rational of using Gauges instead of counters? Unless its expected for the metric to have a decrement. Typically counters are used. 

I'm refereing back to the convo regarding modeling metrics after an http server. Most of those are counters.

op-nat/metrics/metrics.go
```diff
@@ -40,15 +42,51 @@ var (
...
+	})
+
+	acceptanceTestFailed = promauto.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: MetricsNamespace,
+		Name:      "acceptance_test_failed",
+		Help:      "Number of failed acceptance tests",
+	}, []string{
+		"network_name",
+		"run_id",
+	})
```
jelias2: Seems to missing an acceptance test skipped in this file as well

jelias2: It would also be nice to emit metrics with labels suchs as 
gate=<>, suite=<>, test=<>

To all for grouping tests together along with the run_id

Can be backlogged if you like

op-nat/metrics/metrics.go
```diff
@@ -99,20 +137,20 @@ func RecordValidation(network string, runID string, valName string, valType stri
...
+	total int,
+	passed int,
+	failed int,
+	duration time.Duration,
+) {
+	acceptanceResults.WithLabelValues(network, runID, result).Set(1)
+	acceptanceTestTotal.WithLabelValues(network, runID).Set(float64(total))
+	acceptanceTestPassed.WithLabelValues(network, runID).Set(float64(passed))
+	acceptanceTestFailed.WithLabelValues(network, runID).Set(float64(failed))
+	acceptanceTestDuration.WithLabelValues(network, runID).Set(duration.Seconds())
```
jelias2: hmm, to allow for finer metrics. |

I think total failures and total acceptances should come from aggregating individual test results in the grafana backend, rather than aggegrating them in the service 


The query in grafana could look like

sum by (runID) acceptance_test_result{result="passed"}
sum by (runID) acceptance_test_result{result="failed"}

jelias2: I think the metrics need a little tweaking, but I think this can be backlogged, once its pivoted fully to test runner re-work these

op-nat/nat.go
```diff
@@ -3,25 +3,32 @@ package nat
...
-	config  *Config
-	params  map[string]interface{}
-	version string
-	results []ValidatorResult
+	ctx      context.Context
+	config   *Config
+	version  string
+	registry *registry.Registry
+	runner   runner.TestRunner
+	result   *runner.RunnerResult
```
jelias2: Should the logger be apart of the nat object instead of the config object?

op-nat/nat.go
```diff
@@ -85,47 +118,157 @@ func (n *nat) Stopped() bool {
...
+		"",
+	})
+
 	t.Render()
 
 	// Emit metrics
-	// TODO: This shouldn't be here; needs a refactor
-	// TODO: don't hardcode the network name
-	metrics.RecordAcceptance("todo", runID, overallResult.String())
+	metrics.RecordAcceptance(
```
jelias2: Metrics should be emitted as each test case completes instead of when the suite ends to allow for viewing test progression, This may already be the case, I will double check

op-nat/registry/registry.go
```diff
@@ -0,0 +1,228 @@
...
+
+	cfg.Log.Debug("Registry loaded", "len(validators)", len(r.validators))
+
+	return r, nil
+}
+
+// loadValidators loads a test source and its configuration
+func (r *Registry) loadValidators(cfgPath string) error {
+	r.mu.Lock()
+	defer r.mu.Unlock()
```
jelias2: Just curious what the purpose of the mutex is? 

Will keep reviewing... maybe it becomes apparent

op-nat/registry/registry.go
```diff
@@ -0,0 +1,228 @@
...
+	}
+
+	gateMap := make(map[string]types.GateConfig)
+	for _, gate := range config.Gates {
+		gateMap[gate.ID] = gate
+	}
+
+	// Check for circular inheritance before resolving
+	for _, gate := range config.Gates {
+		if err := r.checkCircularInheritance(gate.ID, gate.Inherits, gateMap, make(map[string]bool)); err != nil {
```
jelias2: Nice forward thinking

op-nat/README.md
```diff
@@ -1,54 +1,69 @@
...
-    DefaultParams: MyValidatorParams{
-        // Default parameter values
-    },
-    Fn: func(ctx context.Context, log log.Logger, config nat.Config, _ interface{}) (bool, error) {
-        // Your validation logic here
-        return true, nil
-    },
-}
+### Adding a new test/suite/gate
+All tests, suites, and gates are defined in a `validators.yaml` file. The filename is not important.
```
sigma: is there a (current) requirement for the location of these files?

op-nat/README.md
sigma: it might be slightly tricky. I think we'd need to analyze the file to extract the list of individual test cases (as a test file is not always self-contained, so we might need to load the entire module anyway).
Probably easier to start with specifying individual tests

op-nat/config.go
```diff
@@ -1,55 +1,65 @@
...
 	}
 
-	wallets := []*wallet.Wallet{}
-	for i, w := range manifest.L1.Wallets {
-		w, err := wallet.NewWallet(w.PrivateKey, fmt.Sprintf("user-key-%s", i))
-		if err != nil {
-			log.Warn("error creating wallet: %w", err)
-		}
-		wallets = append(wallets, w)
+	// Parse kurtosis-devnet manifest
```
sigma: nit: it's meant to *not* be kurtosis-specific :)

op-nat/justfile
```diff
@@ -29,7 +29,7 @@ clean:
 # Run tests
 # test: (go_test "./...")
 test:
-    go build ./... && go test -v ./...
+    CGO_ENABLED=0 go build ./... && go test -count=1 -v ./...
```
sigma, line 32: I'm curious: why disable the test cache?

op-nat/nat.go
```diff
@@ -85,47 +118,157 @@ func (n *nat) Stopped() bool {
...
-
-	var overallResult ResultType = ResultPassed
-	var hasSkipped = false
-	var overallErr error
-	for _, res := range n.results {
-		appendResultRows(t, res)
-		overallErr = errors.Join(overallErr)
-		if res.Result == ResultFailed {
-			overallResult = ResultFailed
+	t.SetTitle(fmt.Sprintf("NAT Results (%s)", formatDuration(n.result.Duration)))
```
sigma: It would be nice to separate the raw data from its formatting (I can guarantee we'll want a json output in 3... 2.. 1.. :))

op-nat/runner/runner.go
```diff
@@ -0,0 +1,660 @@
...
+	"strings"
+	"time"
+
+	"github.com/ethereum-optimism/infra/op-nat/registry"
+	"github.com/ethereum-optimism/infra/op-nat/types"
+	"github.com/ethereum/go-ethereum/log"
+)
+
+// TestResult captures the outcome of a single test run
+type TestResult struct {
```
sigma: should we have some support for test artifacts here? logs for  example

op-nat/runner/runner.go
```diff
@@ -0,0 +1,660 @@
...
+	Metadata types.ValidatorMetadata
+	Status   string
```
sigma: I'm a bit worried about leaving the status free-form. Maybe some enum would be appropriate

op-nat/runner/runner.go
```diff
@@ -0,0 +1,660 @@
...
+	Error    string
```
sigma: I'd use the error interface here. At least that'd leave some room for detecting error types

op-nat/runner/runner.go
```diff
@@ -0,0 +1,660 @@
...
+	suiteStart := time.Now()
+	suiteResult := &SuiteResult{
+		ID:    suiteName,
+		Tests: make(map[string]*TestResult),
+		Stats: ResultStats{StartTime: suiteStart},
+	}
+	gateResult.Suites[suiteName] = suiteResult
+
+	// Run all tests in the suite
+	for _, validator := range suiteTests {
```
sigma: not urgent at all, but it'd be nice to have some sort of job executor here, to abstract away the loop and make some room for parallel treatment, limitation of the number of workers, and so on.

op-nat/runner/runner.go
```diff
@@ -0,0 +1,660 @@
...
+	if err != nil {
+		return nil, err
+	}
+	r.log.Info("runAllTestsInPackage() found tests", "package", metadata.Package, "count", len(testNames))
+	return r.runTestList(metadata, testNames)
+}
+
+// listTestsInPackage returns all test names in a package
+func (r *runner) listTestsInPackage(pkg string) ([]string, error) {
+	listCmd := exec.Command("go", "test", pkg, "-list", "^Test")
```
sigma: probably  want to make the location of the "go" binary configurable from main. Someone somewhere (probably a fellow nix user) won't have it in the PATH :)

op-nat/README.md
scharissis: Yes it's non-trivial, otherwise I'd have just added it.
I believe `go test` doesn't facilitate "run this file". IDE plugins, for example, handle this by inspecting the target file and extracting all the test names within. Then passing that list to `go test`. Something like `go test -run ^(TestA|TestB|TestC) x/y/package`

op-nat/config.go
scharissis: I considered that but decided the overhead is trivial. Do you disagree?

They'd add this to fix/debug a single test (say, 'TestFindRPCEndpoints'):
```
gates:
  - id: debug
    tests:
      - name: TestFindRPCEndpoints
        package: github.com/ethereum-optimism/optimism/kurtosis-devnet/pkg/kurtosis/api/run
```

Then run NAT/test-runner with `-gate debug`

op-nat/README.md
scharissis: There is not. Only that there's just one of them.

op-nat/runner/runner.go
```diff
@@ -0,0 +1,660 @@
...
+		Status:   result,
+		Error:    r.formatErrors(errors),
+	}, nil
+}
+
+// runIndividualTest runs a single test and returns its result
+func (r *runner) runIndividualTest(pkg, testName string) (bool, string) {
+	r.log.Debug("Running individual test", "testName", testName)
+
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
```
jelias2: 5 Second timeout seems a bit short, If an l2 has 2s block times, thats only 2 blocks. 

It would be cool if in the validator.yaml the users could configure a timeout

op-nat/runner/runner.go
```diff
@@ -0,0 +1,660 @@
...
+	}
+
+	return &TestResult{
+		Metadata: metadata,
+		Status:   types.TestStatusPass,
+	}, nil
+}
+
+// buildTestArgs constructs the command line arguments for running a test
+func (r *runner) buildTestArgs(metadata types.ValidatorMetadata) []string {
```
jelias2: I thought the idea was to use go test -json?

jelias2 (approved):
Overall the main functionality for test-running logic is there imo. LG(reat)TM

op-nat/README.md
scharissis: Added this to an idea backlog.

op-nat/README.md
scharissis: Added this to an idea backlog.

sigma (approved):

op-nat/config.go
scharissis: Absolutely. Removed in d808c28.
Thanks!

op-nat/example-validators.yaml
scharissis: Not as-is. Do you think we'll need this?
My reasoning for not needing this is that if we don't "force" suites to be part of gates then they could become stale/unused over time.

eg.:
```
- suite1: [test1]
- suite2: [test2]
- suite3: [test3]

- gate1: [suite2, suite3]
```
In this example suite1 and test1 are unused (and thus useless?) (and we may not notice among all the config)

op-nat/metrics/metrics.go
scharissis: I keep going back and forth on this myself. Can always reassess when we're actually consuming them.
My thinking is essentially:
- 'totals' should be Counters (number of tests only increases)
- 'results' perhaps should be Gauges (result of each test goes "up & down" between 0,1 [fail, pass]

op-nat/config.go
jelias2: Good with me

op-nat/example-validators.yaml
jelias2: Yeah I get that, I feel as though for composability purposes it may be worth it. 

```
- suite1: [test1]
- suite2: [test2]
- suite3: [test3]

- gate1: [suite2, suite3]
- gate2: [suite1, suite3]
```

It would remove suite3 from being defined multiple times. Especially as the number of gates and test grow. 

Thinking similar like ansible with task, plays and playbooks. 

Curios what @sigma thinks

scharissis: Oh! I slightly misunderstood you earlier.

Again, this isn't possible as-is. But we could make it so.

I believe this could be a great improvement, iff.,  we don't end up using gates such that each is a superset of the previous.
But I'm not sure how the gates will be used right now. My guess right now is that they'll look like this:

```
alphanet: [stuff]
betanet: [alphanet, more-stuff-only-for-betanet]
mainnet: [betanet, more-stuff-only-for-mainnet]
```
In other words, each "stricter" gate MUST include all previous gates' tests/criteria.

It's also worth noting that defining suites/tests outside of gates, for more modularity, would almost definitely be better in the general case of a test-runner.

So this really should be thought about more. The only question is, do we need to think about it in this PR/change? I would argue not.

Thoughts? @jelias2 & @sigma ?

op-nat/metrics/metrics.go
scharissis: Created https://github.com/ethereum-optimism/infra/issues/194 to address this properly.
Thanks!

op-nat/metrics/metrics.go
scharissis: On skipped validator metrics, created: [Issue #195 created](https://github.com/ethereum-optimism/infra/issues/195)

On more labels/dimensions, created: [Issue #196 created](https://github.com/ethereum-optimism/infra/issues/196)

op-nat/metrics/metrics.go
scharissis: Let's backlog and reconsider how we approach it.

I was treating it like http_requests_total in the prometheus docs here: https://prometheus.io/docs/practices/naming/#labels

Created the following issue to address it: https://github.com/ethereum-optimism/infra/issues/197

op-nat/nat.go
scharissis: Agreed. I think that's what we have. But there's a different metric (and helper function) for each.
`RecordValidation` is used for tests and `RecordAcceptance` for overall run pass/fail.

op-nat/registry/registry.go
scharissis: Not strictly needed; this is possibly premature optimisation.

It guards the `registry.validators` slice in case of concurrent access, for example multiple calls to `loadValidators`. However, as-is, I don't think that'll happen in practise.

op-nat/registry/registry.go
scharissis: Thanks. And if we add the 'modular suite definitions' idea you mentioned we should probably add a similar check for unused & duplicate validator definitions.

op-nat/config.go
scharissis: Definitely. I've since removed this actually. The assumption now is that the tests _must_ use `devnet-sdk` and it will handle the manifest-parsing.

op-nat/justfile
scharissis: While developing/testing it made the output more predictable and nice.
Then that got me wondering whether or not we should trust the cache or always re-run. Because while the test may remain the same, the system may not.

op-nat/nat.go
scharissis: Isn't it already though? This function, `printResultsTable`, simply iterates over `nat.result` (the data). We could write a similar function `printJSON` by iterating over the same results data.
Or am I missing something?

op-nat/justfile
scharissis: Paranoia :)
I do plan to remove this.

op-nat/runner/runner.go
scharissis: Yes, but I was thinking of tackling this in a separate PR. May be non-trivial to parse it nicely.
https://github.com/ethereum-optimism/infra/issues/198 created to capture the work; please add any more ideas/context to that ticket.

Sound OK?

op-nat/runner/runner.go
scharissis: Agreed. Updated this to an enum (and moved it to the `types` package)

op-nat/runner/runner.go
scharissis: Agreed; updated.

op-nat/runner/runner.go
scharissis: Done; made it configurable (via a flag or envvar).

op-nat/runner/runner.go
scharissis: Changed to 5 minutes.
A per-test timeout would indeed be good. I'll add that to the backlog. Parameterisation in general needs to be reconsidered with this pivot.

op-nat/runner/runner.go
scharissis: it was/is. I forgot while implementing it. :/
I've added a backlog item to revisit this if we don't mind.

PR #203, opened 2025-02-21
op-signer: opsigner_signBlockPayloadV2, fixes, cleanup

protolambda:
**Description**

- Remove the outdated op-signer `client` package. We have the exact same code, but newer and better, as improved in op-node over time, in the monorepo.
  - The old client package duplication comes from a pre-bedrock time where the op-signer code lived in a private repo and wasn't integrated into the monorepo for remote tx signing yet.
- Fix the CLI client sub-command for block-signing. The block-signing-attribute that was sent via the outdated client binding didn't seem right at all (hash instead of block-args). So I am not sure if this test command was ever used. This PR fixes the client sub-command.
- Introduce the `opsigner_signBlockPayloadV2` endpoint, that uses the improved RPC typing. Ideally we document the V2 API, this is good to standardize for alternative OP-Stack implementations. Rust implementations shouldn't have to mimic the previous bad V1 Go JSON typing issues.
- Updated to latest op-geth dependency. May also be useful for the Isthmus workstream. But does not fully support 7702 yet, since this PR focuses on block-signing fixes.

This PR depends on this monorepo PR: https://github.com/ethereum-optimism/optimism/pull/14478

The `go.mod` points to a commit from that PR, and will have to be updated to a canonical commit of the monorepo after the PR is squash-merged.
Do not merge this PR until the above `go.mod` update.

**Tests**

Updated the block-signing tests to also cover the V2 variant.

**Additional context**

Block-signing is part of the sequencer stack, which I am working on improving, and building test tooling with.

**Metadata**

- Fixes #[Link to Issue]


sebastianst:
Does this provide L1 Pectra readiness, or had this already been addressed in op-signer?

protolambda:
> Does this provide L1 Pectra readiness, or had this already been addressed in op-signer?

It updates the geth dep, but doesn't support signing of 7702 txs yet, since the auth-list attribute is not part of the API type yet.
We can add that in a second PR, this PR is focused on block-signing, which is agnostic of hardforks (no full block payload included)

op-signer/service/service.go
```diff
@@ -178,37 +192,39 @@ func (s *OpsignerSerivce) SignBlockPayload(ctx context.Context, args signer.Bloc
...
-	if err := args.Check(); err != nil {
+	msg, err := getMsg()
+	if err != nil {
 		s.logger.Warn("invalid signing arguments", "err", err)
 		labels["error"] = "invalid_blockPayload"
 		return nil, &InvalidBlockPayloadError{message: err.Error()}
 	}
 
-	if *args.SenderAddress != authConfig.FromAddress {
+	if fromAddress != nil && *fromAddress != authConfig.FromAddress {
```
mininny, line 202: If the `fromAddress` field is optional from the op-signer user (op-node) side, should it also be optional in op-signer server's configuration? 

As I understand the code, op-node will always (at least for v1 param for now) fill in the `SenderAddress` field. Then, the user also must specify the fromAddress in the signer server's `authConfig`. 

What do you think about think doing equality check only if both of them exists so that the signer client and server can both freely set the fromAddress?

op-signer/app.go
```diff
@@ -226,47 +237,64 @@ func ClientSign(version string, action SignActionType) func(cliCtx *cli.Context)
...
+			var payloadHash common.Hash
+			if err := payloadHash.UnmarshalText([]byte(payloadHashStr)); err != nil {
+				return fmt.Errorf("failed to unmarshal block payload-hash argument: %w", err)
 			}
-
-			signingHash := common.Hash{}
-			if err := signingHash.UnmarshalText([]byte(blockPayloadHash)); err != nil {
-				return fmt.Errorf("failed to unmarshal block payload argument: %w", err)
+			var chainID eth.ChainID
+			if err := payloadHash.UnmarshalText([]byte(chainIDStr)); err != nil {
```
mininny:
```suggestion
			if err := chainID.UnmarshalText([]byte(chainIDStr)); err != nil {
```

op-signer/app.go
```diff
@@ -226,47 +237,64 @@ func ClientSign(version string, action SignActionType) func(cliCtx *cli.Context)
...
+				return fmt.Errorf("failed to unmarshal block chain-ID argument: %w", err)
+			}
+			var domain eth.Bytes32
+			if err := payloadHash.UnmarshalText([]byte(domainStr)); err != nil {
```
mininny:
```suggestion
			if err := domain.UnmarshalText([]byte(domainStr)); err != nil {
```

mininny (commented):
I did some simple tests on my local machine with the new/old op-signer and V1/V2 rpcs seem to work. Awesome clean up 👍

op-signer/service/service.go
protolambda: - The server value serves as default if the client value is unset
- The client value serves as request if the server supports multiple different signers

~~I believe you're right and we need a nil-check here, good catch.~~ 
Edit: actually, nil check is already there. I think I understand you know, you're suggesting to make it entirely optional on the server side? I think that's not possible, since the server still needs the private key, so that type of config is always there. The client-side is already optional. The `GetAuthConfigForClient` checks that it's a known allowed address, if non-nil.

protolambda:
@mininny thank you for the review! I fixed the client test command, commented on the other issue, and updated the `go.mod`. Ready to merge if you agree on the remaining open comment.

op-signer/service/service.go
mininny: Sorry if it doesn't make sense - it's hard to organize the thoughts. But I was thinking that we could have it optional on server side since the address associated with the private key is fixed and sort of duplicative as the address and private key identify the same thing. Like, the `fromAddress` option is unique for the keyname and chainID. And in most cases, only one account will be used for block signing. So if the client-side is able to skip fromAddress (as it's optional in V2), then we could also have server-side run without the fromAddress config as well. Like `fromAddress != nil && authConfig.FromAddress && *fromAddress != authConfig.FromAddress`

But again, I think it's fine to leave the server to always have fromAddress authConfig, so that the clients can be added/modified without the server changing, and for additional verification.

protolambda: I think it's good practice to sanity-check the sender address, to never accidentally sign with the wrong account. It's easy to misconfigure something in infrastructure otherwise. Let's leave that in place, I think it's useful.

mininny: Sounds good! Thanks :+1:

mininny (approved):

zhwrd (approved):

PR #209, opened 2025-02-27
Use signBlockPayload metric properly

mininny:
**Description**
op-signer was incorrectly using `signer_signtransaction_total` when signing block payload. Create a new prometheus metric `signer_signBlockPayload_total` and use that when signing block payload.


jelias2 (approved):

jelias2:
Does the new metric need to be registered?  See [link](https://github.com/ethereum-optimism/infra/blob/2b7296456e592a99487766bc77fd9000b400301d/op-signer/app.go?plain=1#L95)

mininny:
> Does the new metric need to be registered? See [link](https://github.com/ethereum-optimism/infra/blob/2b7296456e592a99487766bc77fd9000b400301d/op-signer/app.go?plain=1#L95)

You're right. Thanks for flagging!

PR #216, opened 2025-03-05
feat: adjust metric to include multicall in rpc_requests_total metric

jelias2:
**Description**
* Previously multicall would not increment rpcRequestsTotal which resulted in odd metric behavior, adjusting this to be incremented prior to calling ExecuteMulticall
**Tests**

**Additional context**

**Metadata**

- Fixes #[Link to Issue]

### DIFF
--- a/proxyd/backend.go
+++ b/proxyd/backend.go
@@ -780,14 +780,14 @@ func (bg *BackendGroup) Forward(ctx context.Context, rpcReqs []*RPCReq, isBatch 
 		rpcReqs, overriddenResponses = bg.OverwriteConsensusResponses(rpcReqs, overriddenResponses, rewrittenReqs)
 	}
 
+	rpcRequestsTotal.Inc()
+
 	// When routing_strategy is set to 'multicall' the request will be forward to all backends
 	// and return the first successful response
 	if bg.GetRoutingStrategy() == MulticallRoutingStrategy && isValidMulticallTx(rpcReqs) && !isBatch {
 		backendResp := bg.ExecuteMulticall(ctx, rpcReqs)
 		return backendResp.RPCRes, backendResp.ServedBy, backendResp.error
 	}
-
-	rpcRequestsTotal.Inc()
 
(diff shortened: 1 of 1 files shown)


zhwrd (approved):

zhwrd:
make sense, but because the multi-call metric is emitted N more times than rpc_req_total, isnt it still likely to hit this error if one of the backends is down?

proxyd/backend.go
```diff
@@ -780,15 +780,15 @@ func (bg *BackendGroup) Forward(ctx context.Context, rpcReqs []*RPCReq, isBatch
 		rpcReqs, overriddenResponses = bg.OverwriteConsensusResponses(rpcReqs, overriddenResponses, rewrittenReqs)
 	}
 
+	rpcRequestsTotal.Inc()
```
scharissis, line 783: for extra safety, should we defer this call?

```suggestion
	defer rpcRequestsTotal.Inc()
```

scharissis (approved):

jelias2:
> make sense, but because the multi-call metric is emitted N more times than rpc_req_total, isnt it still likely to hit this error if one of the backends is down?

If I understand what your getting at, each backend could fail, would would increment unserviceable_requests N times, and rpc_requests_total is only incremented 1 time per request.

So this could still result is odd alerting scenarios. Since there could be more unservicable_requests than total_requests


There aren't any multicall specific metrics or alerts yet, this is definetly a gap to improve on 
However even in normal non-multicall world, it could be the case where 1 client request fails each of the backends. which could still causes this alerting behavior of more failures


Overall I think the metrics could be improved by honing in a distinctions between client_request and backend_requests

For 1 client request there can be N*M backend requests. N being the number of backends and M being the amount of retries per backend.

proxyd/backend.go
jelias2: I don't have a major preference. I guess the only concern would be deferring the metric call could make metrics lag a bit. Maybe it could take an extra 5-10s for the metric ex if all the requests timeout in 5s

PR #218, opened 2025-03-06
[op-acceptor] periodic test running

### LINKED ISSUE #136: [nat] Run on a schedule

scharissis:
**Description**

Tests run continuously on a configurable schedule.

**Tests**

Added some high-level NAT tests to ensure things start/stop/run as expected.

**Metadata**

- Supersedes #166 
- Closes #136

### DIFF
--- a/op-acceptor/config.go
+++ b/op-acceptor/config.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"path/filepath"
+	"time"
 
 	"github.com/urfave/cli/v2"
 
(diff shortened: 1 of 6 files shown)


janjakubnanista (commented):
Do we need the interval functionality built into the runner? As opposed to an external solution (crontab or scheduled pipelines)?

scharissis:
Thanks for the review, @janjakubnanista 

> Do we need the interval functionality built into the runner? As opposed to an external solution (crontab or scheduled pipelines)?

Good question. Either approach _should_ work.
I slightly prefer this in-code approach since:
1. It works for all deployment avenues (local, CI, k8s)
2. It's the same code/implementation for all deployment avenues
3. I'd rather modify Go

Did you have some concerns? Let me know if you wanted to chat further about it.

janjakubnanista (approved):

PR #224, opened 2025-03-07
feat(proxyd): add config to skip is syncing check

hmvien:
**Description**

I'm using proxyd as a generalize purpose consensus aware LB for nodes. In some configuration, checking for the `eth_syncing` state of the node is not ideal. For instance, for Orbit's nitro nodes, it is more important to have an active sequencer feed for the nodes than to check for the sync state between the node and the parent chain. For this use-case, I would like to configure proxyd to disable this check. Ideally I should add specific handling for nitro nodes, but that's probably outside the scope of proxyd. 

**Tests**

I ran proxyd without this config and confirmed syncing state is still checked. I ran proxyd with the config and confirmed syncing state is not checked.

### DIFF
--- a/proxyd/backend.go
+++ b/proxyd/backend.go
@@ -151,6 +151,7 @@ type Backend struct {
 	stripTrailingXFF     bool
 	proxydIP             string
 
+	skipIsSyncingCheck bool
 	skipPeerCountCheck bool
 	forcedCandidate    bool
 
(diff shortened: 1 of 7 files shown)


jelias2 (commented):
Overall LGTM, would you mind adding a test case for this behavior for approving for merge?

hmvien:
@jelias2 thanks for reviewing. I added test. Mind taking another look when you get a chance?

jelias2 (approved):

jelias2:
Thanks for the contribution Hien! The latest relase: proxyd/v4.14.3 has this feature.

PR #229, opened 2025-03-13
feat(proxyd): add threshold configs to handle safe and finalized block drift

hmvien:
**Description**

Added two new configs: `safe_block_drift_threshold` and `finalized_block_drift_threshold` with default value of 0. These configs compliments the safe and finalized heads check [here](https://github.com/ethereum-optimism/infra/blob/e10211331a1114607227ea0acdc40ed7f5568b81/proxyd/consensus_poller.go#L409). By default, the check will still ban backends that are experiencing consistency issues. We found this behavior to not work well for some L2s RPCs where drifts happen very often. 

We observed RPC providers prioritized having unsafe head consistency over safe and finalized heads consistency. This is fine for our use-case and we would like to tune our proxyd instances to increase tolerance for inconsistency. 

To be clear, this change does not impact existing deployment of proxyd. Developers will need to opt into this config to relax the tolerance level.

**Tests**

I added integration tests to make sure backends are not being banned if the inconsistency range is within tolerance. They should be banned if it is above the tolerance.

### DIFF
--- a/proxyd/integration_tests/consensus_custom_config_test.go
+++ b/proxyd/integration_tests/consensus_custom_config_test.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/ethereum-optimism/infra/proxyd"
 	ms "github.com/ethereum-optimism/infra/proxyd/tools/mockserver/handler"
+	"github.com/ethereum/go-ethereum/common/hexutil"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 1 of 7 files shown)


proxyd/integration_tests/consensus_custom_config_test.go
```diff
@@ -129,3 +130,159 @@ func TestConsensusSkipSyncTest(t *testing.T) {
 		require.Equal(t, 2, len(consensusGroup))
 	})
 }
+
```
yashvardhan-kukreja, line 133: Small thing:

Can we have something like 
```go
	initSetupAndAssetions := func() {
		reset()
		update()
		
		require.Equal(t, "0x101", bg.Consensus.GetLatestBlockNumber().String())
		require.Equal(t, "0xe1", bg.Consensus.GetSafeBlockNumber().String())
		require.Equal(t, "0xc1", bg.Consensus.GetFinalizedBlockNumber().String())
	}
```
and use this as the first line under every t.Run()?

(improves the readability, ensures that every subtest starts from a fresh slate independent of the result of the previous sub-test)

yashvardhan-kukreja:
@hmvien Just the above comment, rest all looks good to me.

hmvien:
@yashvardhan-kukreja thanks for the suggestion, I committed the changes. Mind taking another look when you get a chance? :)

jelias2:
Have you taken a look at the Consensus Poller Max Block Lag Flag? That flag will allow backend to still be included as candidates even if they fall behind consensus. 

CheckBlockTags is typically used for making sure that backend are always increasing with block height. The reason the ban is zero blocks is because decrementing block height is a more of an outlier/reorg than purley lagging, which is the reason to ban it immediately


Just curious on the behavior, if a backend is lagging or actually reverting on its height. Feel free to post logs or context

hmvien:
Hey @jelias2,

We're observing RPC reverting on their safe and finalized heights. We don't see this happening much on Ethereum, but see it happen almost all the time for Base using Quicknode's RPC. This is actually blocking us from adding more RPC providers to act as fallbacks in case our primary RPC from Alchemy fail. Our theory is that Base or any chains with a very low block time will cause problems. We have seen block reverting by >200 blocks.

image
image

We understand the risk adding this tolerance, but we feel the consensus checks to ensure no backends can be too far behind will protect us from serving RPCs that have halted.

jelias2:
Hey @hmvien overall LGTM thanks for sharing the logs. 

Do you mind updating the function header comments / documenting the new conditionals in checkExpectedBlockTags and then should be good to merge?

hmvien:
@jelias2 updated

jelias2 (approved):

jelias2:
The changes are now in `proxyd/v4.14.4`. Thanks for contributing!

PR #231, opened 2025-03-14
feat(op-acceptor): Introduce run-once mode for CI use-cases, enable by default

teddyknox:
## Summary
This PR changes the default behavior of op-acceptor to run tests once and exit (run-once mode), rather than running continuously. The existing continuous mode is preserved by explicitly setting `--run-interval`.

## Motivation
In a CI or manual testing environment users want to run tests once, get results, and exit. The current default requires manual termination, which is inconvenient for those use-cases.

## Changes
- Changed default `--run-interval` from `1h` to `0` (run-once mode)
- Added `RunOnce` config field to track execution mode
- Modified `Start()` to handle run-once mode and trigger application shutdown
- Updated documentation and added test case

## Backward Compatibility
Users who need continuous execution can specify `--run-interval=1h` to maintain the same behavior.

## Testing
Implementation verified with unit tests, build validation, and linting.

### DIFF
--- a/op-acceptor/README.md
+++ b/op-acceptor/README.md
@@ -91,6 +91,17 @@ go run cmd/main.go \
   --validators validators.yaml \    # Path to the validator definitions
 ```
 
+By default, op-acceptor will run tests once and then exit, which is ideal for CI/CD pipelines and one-off testing.
+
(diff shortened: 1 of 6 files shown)


scharissis (approved):
This is great - thank you!

PR #234, opened 2025-03-15
Set Exit Status Codes for op-acceptor run-once mode

teddyknox:
This is necessary for using op-acceptor in CI, where the exit code is the easiest way to tell the caller whether the test job should succeed or not.

## Description
Enhances op-acceptor to return meaningful exit status codes in run-once mode:
- **0**: All tests pass or skip (success)
- **1**: One or more tests fail (test failure)
- **2**: Runtime errors occur (system failure)

Added tests to verify correct behavior.

This makes it easier to use op-acceptor in CI/CD pipelines where exit codes determine workflow outcomes.

### DIFF
--- a/op-acceptor/cmd/main_test.go
+++ b/op-acceptor/cmd/main_test.go
@@ -1,0 +1,258 @@
+package main_test
+
+import (
+	"bytes"
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"testing"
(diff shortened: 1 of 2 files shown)


scharissis (approved):
Awesome, thank you!

scharissis:
This implements https://github.com/ethereum-optimism/infra/issues/236

Issue #241, opened 2025-03-17
[acceptance] Handle test panics


scharissis:
We should safeguard our acceptance test running against tests which panic so that one "bad" test doesn't prevent the test suite from completing.

scharissis:
Done in https://github.com/ethereum-optimism/infra/pull/242

PR #242, opened 2025-03-17
feat(op-acceptor): handle test panics

scharissis:
**Description**

* Handle potential test panics to safeguard the remainder of the tests
* Defines the CLIs exit codes and reuses them throughout main and tests
* Simplifies some of the testing code in which we create go packages & code

**Metadata**

- Implements https://github.com/ethereum-optimism/infra/issues/241

### DIFF
--- a/op-acceptor/cmd/main_test.go
+++ b/op-acceptor/cmd/main_test.go
@@ -2,60 +2,37 @@ package main_test
 
 import (
 	"bytes"
+	"context"
 	"fmt"
 	"os"
 	"os/exec"
 	"path/filepath"
 	"testing"
+	"time"
 
+	"github.com/ethereum-optimism/infra/op-acceptor/exitcodes"
 	"github.com/stretchr/testify/require"
 )
 
 // TestExitCodeBehavior verifies that op-acceptor returns the correct exit codes in run-once mode:
 // - Exit code 0 when all tests pass
 // - Exit code 1 when any tests fail
-// - Exit code 2 when there's a runtime error (panic)
+// - Exit code 2 when there's a runtime error
 func TestExitCodeBehavior(t *testing.T) {
 	// Setup paths
 	cwd, err := os.Getwd()
 	require.NoError(t, err, "Failed to get current working directory")
-
 	projectRoot := filepath.Dir(cwd)
 
 	// Binary path
 	opAcceptorBin := filepath.Join(projectRoot, "bin", "op-acceptor")
-
-	// Build the binary if it doesn't exist
-	if !fileExists(opAcceptorBin) {
-		t.Logf("Building op-acceptor binary...")
-
-		// Create bin directory if needed
-		err = os.MkdirAll(filepath.Dir(opAcceptorBin), 0755)
-		require.NoError(t, err, "Failed to create directory for binary")
-
-		// Build the binary
-		buildCmd := exec.Command("go", "build", "-o", opAcceptorBin, filepath.Join(projectRoot, "cmd"))
-		var buildOutput bytes.Buffer
-		buildCmd.Stdout = &buildOutput
-		buildCmd.Stderr = &buildOutput
-
-		err = buildCmd.Run()
-		if err != nil {
-			t.Logf("Build output:\n%s", buildOutput.String())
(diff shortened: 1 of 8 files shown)


op-acceptor/gandalf.go
```diff
@@ -1,9 +1,16 @@
...
-import "fmt"
+import (
+	"fmt"
+	"testing"
+)
 
 // printGandalf gatekeeps naughty network deployments
 func printGandalf() {
+	if testing.Testing() {
+		// skip when running tests
```
teddyknox, line 11: Nice

op-acceptor/nat.go
```diff
@@ -85,6 +86,14 @@ func New(ctx context.Context, config *Config, version string, shutdownCallback f
 // Start runs the acceptance tests periodically at the configured interval.
 // Start implements the cliapp.Lifecycle interface.
 func (n *nat) Start(ctx context.Context) error {
+	// Set up panic recovery to ensure we exit with code 2 for runtime errors
+	defer func() {
+		if r := recover(); r != nil {
+			n.config.Log.Error("Runtime error occurred", "error", r)
+			os.Exit(exitcodes.RuntimeErr)
```
teddyknox, line 93: We're not putting this in main.go because that's a different goroutine, correct?

op-acceptor/nat.go
```diff
@@ -102,25 +111,18 @@ func (n *nat) Start(ctx context.Context) error {
 	// Run tests immediately on startup
 	err := n.runTests()
 	if err != nil {
-		// For other errors, return an ExitCoder with exit code 1
-		n.config.Log.Error("Error running tests", "error", err)
-		return cli.Exit(fmt.Sprintf("Error running tests: %v", err), 2)
+		if strings.Contains(err.Error(), "test failures:") {
```
teddyknox: Probably better to define a specific type for a test error and then do a type query here than inspect the error string.

op-acceptor/nat.go
```diff
@@ -168,17 +170,23 @@ func (n *nat) Start(ctx context.Context) error {
...
+
+	// Handle runtime errors (result will be nil or incomplete)
 	if err != nil {
 		n.config.Log.Error("Error running tests", "error", err)
-		return err
+		return fmt.Errorf("runtime error: %w", err)
 	}
-	n.result = result
 
+	n.result = result
```
teddyknox, line 180: quite imperative style to set this rather than return it

teddyknox (approved):

op-acceptor/nat.go
teddyknox: The fact that we have two different `os.Exit(exitcodes.RuntimeErr)` calls reveals the complexity introduced by doing the initial test run within `Start()`... not a huge deal though.

scharissis: Agreed.
Let me know if you have any ideas on reducing this complexity.

op-acceptor/nat.go
scharissis: Agreed. Will do this in a follow-up PR so we can minimise the open branches to resolve this weird issue in CI we're seeing right now. Thanks

op-acceptor/cmd/main_test.go
```diff
@@ -123,63 +118,91 @@ func TestExitCodeBehavior(t *testing.T) {
...
+			t.Logf("Build output:\n%s", buildOutput.String())
+			t.Fatalf("Failed to build op-acceptor binary: %v", err)
+		}
 
-	require.NoError(t, os.WriteFile(goModPath, []byte(goModContent), 0644))
+		t.Logf("Successfully built binary at %s", binaryPath)
+	}
+
+	// Verify binary exists
+	require.FileExists(t, binaryPath, "op-acceptor binary not found")
```
janjakubnanista, line 147: Maybe a tad stronger verification - check if file is executable

janjakubnanista (approved):
One small nitpick but approving nevertheless

op-acceptor/cmd/main_test.go
scharissis: Good thinking. I'll do this in a follow-up PR so this doesn't block downstream work. Thanks!

PR #248, opened 2025-03-20
feat(op-acceptor): use json output for go tests.

### LINKED ISSUE #180: [acceptance] JSON output parsing
To increase robustness, parse the test JSON output (from go-test).

scharissis:
**Description**

When invoking `go test` to run the acceptance tests, always add the `-json` flag to get JSON output, which is more reliable to parse.

**Metadata**

- Fixes https://github.com/ethereum-optimism/infra/issues/180

### DIFF
--- a/op-acceptor/cmd/main_test.go
+++ b/op-acceptor/cmd/main_test.go
@@ -14,25 +14,31 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+const (
+	passing_dir = "passing"
+	failing_dir = "failing"
+	panic_dir   = "panicking"
+)
+
 // TestExitCodeBehavior verifies that op-acceptor returns the correct exit codes in run-once mode:
 // - Exit code 0 when all tests pass
 // - Exit code 1 when any tests fail
 // - Exit code 2 when there's a runtime error
 func TestExitCodeBehavior(t *testing.T) {
-	// Setup paths
-	cwd, err := os.Getwd()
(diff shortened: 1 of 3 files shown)


teddyknox (approved):

op-acceptor/cmd/main_test.go
```diff
@@ -151,46 +154,58 @@ func ensureBinaryExists(t *testing.T, projectRoot, binaryPath string) {
...
 
-import "testing"
+	// Create a go.mod file to make this a valid module
+	goModPath := filepath.Join(packageDir, "go.mod")
+	goModContent := `module test
 
-func %s(t *testing.T) {
-	%s
-}
+go 1.21
```
janjakubnanista, line 174: Why not just add a `testdata` directory? It would spare you the trouble of having to dynamically create files

scharissis: I like the flexibility of this approach.
Open to changing it to `testdata`, but I don't have a compelling reason right now.

sigma: note that for the more  dynamic cases, using `T.TempDir()` as a starting point is probably preferable, as it's unique and cleaned up automatically

sigma (approved):

scharissis:
> note that for the more dynamic cases, using T.TempDir() as a starting point is probably preferable, as it's unique and cleaned up automatically

Thanks. This is actually what I originally had, but it wasn't playing nice with go modules!

PR #122, opened 2025-01-15
feat(op-signer): add support for AWS KMS provider

ddaws:
**Description**

This PR adds AWS KMS support to the op-signer service, allowing users to choose between AWS KMS and Google Cloud KMS for signing operations. The implementation maintains backward compatibility by defaulting to GCP if no provider is specified.

```yaml
provider: "AWS" # or "GCP" (defaults to "GCP" if not specified)
auth:
  - name: "test-client"
    key: "test-key"
```

**Changes**

- Add provider type configuration in YAML config
- Add tests for config parsing and validation of new provider type
- Create provider factory pattern for KMS provider selection
- Update service initialization to handle provider errors
- Add AWS KMS signature provider implementation
- Add AWS SDK dependencies

**Breaking Changes**

None. Existing configurations without a provider specified will continue to use GCP KMS.

**Tests**

Config Tests (op-signer/service/config_test.go):

1. TestReadConfig_DefaultKeyProvider
    - Tests that when no provider is specified in the config, it defaults to GCP
    - Creates a temporary config file without a provider field
    - Verifies that config.ProviderType equals provider.KeyProviderGCP
1. TestReadConfig_ExplicitKeyProvider
    - Tests that when AWS is explicitly specified as the provider, it's properly parsed
    - Creates a temporary config file with provider: "AWS"
    - Verifies that config.ProviderType equals provider.KeyProviderAWS
1. TestReadConfig_InvalidKeyProvider
    - Tests that an invalid provider type results in an error
    - Creates a temporary config file with provider: "INVALID"
    - Verifies that an error is returned containing "invalid provider"

AWS KMS Tests (op-signer/service/provider/aws_kms_test.go):

1. TestAWSKMSSignatureProvider_GetPublicKey
    - Tests public key retrieval from AWS KMS
1. TestAWSKMSSignatureProvider_SignDigest
    - Tests signature generation and verification
    - Verifies 65-byte signature length
    - Tests signature recoverability
    - Verifies recovered public key matches original
    - Validates signature using go-ethereum's implementation

Utility Tests (op-signer/service/provider/utils_test.go):

1. TestMarshalAndUnmarshalECDSAPublicKey
    - Tests secp256k1 public key DER encoding/decoding
1. TestMarshalAndParseECDSAPublicKey
    - Validates compatibility with AWS KMS format
1. TestConvertEthereumSignatureToDER
    - Tests signature format conversion
1. TestConvertToEthereumSignatureRecoverable: Tests signature recoverability
[...]

PR #252, opened 2025-03-20
feat(op-signer): local KMS provider

edobry:
## Description

This PR implements a new `LOCAL` KMS provider in order to enable signing transactions using a local private key file, intended for use only in local development and testing environments.

## Changes

This PR:
- pulls the `provider` package up to the same level as `service`
- renames `SignerServiceConfig` to `ProviderConfig` and moves it to the `provider` package
- renames `CloudKMSSignatureProvider` to `GCPKMSSignatureProvider`
- implements the `LocalKMSSignatureProvider`
- updates the `README` to document the configuration format
- updates the `gen-local-tls.sh` script

## Testing

To test this feature, you can use the following steps:

1. generate a private key

```shell
openssl ecparam -name secp256k1 -genkey -noout -param_enc explicit -out "ec_private.pem"
```

2. configure `op-signer` to use this key

```yaml
provider: LOCAL
auth:
  - name: localhost
    key: ec_private.pem
```

3. use the `./gen-local-tls.sh` script to generate a self-signed CA and mTLS keys
4. trust the CA root on your machine (MacOS only)

```shell
sudo security add-trusted-cert -d -r trustRoot -k /Library/Keychains/System.keychain tls/ca.crt
```

5. build & run `op-signer`

```shell
make build && ./bin/signer
```

6. send a test RPC

```shell
curl -X POST -H  "Content-Type: application/json" \
  --cert tls/tls.crt \
  --key tls/tls.key \
  -d @test-rpc.json \
  https://localhost:8080
```

7. confirm that the first TX succeeds and the second fails

```json
[
  {
    "jsonrpc": "2.0",
    "id": "1",
    "result": "0x02f89d82053980010182753094000000000000000000000000000000000000aaaa0180f838f794000000000000000000000000000000000000aaaae1a0000000000000000000000000000000000000000000000000000000000000000080a0ba1bf0d4e091881d59bd0f38f16a99906d7dfb33a2e298e04162b863672d1504a07d55d98ace9950ec23a1a0da02bb8dc497f7419e0057a930119d3e6769e61c77"
  },
  {
    "jsonrpc": "2.0",
    "id": "1",
    "error": {
      "code": -32010,
      "message": "nonce not specified"
    }
  }
]
```

## Context

> This PR currently targets the AWS KMS branch (#122) in order to simplify the diff, and needs to be repointed at `main` once that PR is merged.

In service of: https://github.com/ethereum-optimism/platforms-team/issues/581


op-signer/gen-local-tls.sh
```diff
@@ -7,39 +7,51 @@ TLS_DIR=$SCRIPT_DIR/tls
 
 version=$(openssl version)
 
-if [[ "$version" != "LibreSSL"* ]] && [[ "$version" != "OpenSSL 1.1"* ]]; then
```
ddaws, line 10: I've put up a PR to change `gen-local-tls.sh` to use the `alpine/openssl` docker image. This should help make this script more portable, so we don't need to depend on specific versions of installed TLS libs

Here is the PR https://github.com/ethereum-optimism/infra/pull/255

op-signer/gen-local-tls.sh
```diff
@@ -7,39 +7,51 @@ TLS_DIR=$SCRIPT_DIR/tls
...
+  echo "minimum required version: $minimumVersion"
+  echo "please upgrade OpenSSL to >= $minimumVersion"
+  exit 1
+fi
+
+echo -e "Generating mTLS credentials for local development...\n"
 
 mkdir -p "$TLS_DIR"
 
+org="OP Labs"
```
jelias2, line 30: Should the org be something more generic? Like `OP-Signer Local Org`?

op-signer/provider/local_kms.go
jelias2, line 1: Nit: the functions here could benefit from a bit more logging imo, maybe debug level if you want to keep the logs clean

jelias2 (approved):

op-signer/gen-local-tls.sh
edobry: perhaps, I just kept it as-is. its only used in testing anyway

op-signer/provider/local_kms.go
edobry: yeah I can add more logging, what sort of things do you think would be relevant?

zhwrd:
merged the upstream pr, can you rebase onto main?

op-signer/gen-local-tls.sh
edobry: that seems like a good change, i'll review, ty! i'm about to merge this PR however, so we'll need to update your PR once that happens to account for my changes. happy to take that task

op-signer/gen-local-tls.sh
ddaws: Update to `OP-Signer Local Org` here:
 https://github.com/ethereum-optimism/infra/pull/255/files#diff-48ea7eacdc0406935a5d46703b44e6014f4eaa35126ee1d6d66a030b3ecd91ceR13

op-signer/gen-local-tls.sh
ddaws: Thanks @edobry. I've fixed merge conflicts #255

PR #262, opened 2025-03-24
Add timeout configuration support

### LINKED ISSUE #261: Add configurable timeout support to op-acceptor
Useful for supporting long-running NAT tests.

teddyknox:
**Description**

Adds the ability to configure the timeout of an individual test or test package, and also adds the ability to update the default test timeout for an op-acceptor startup invocation via CLI.

**Tests**

I've tested the command building.

**Additional context**

**Metadata**

- Fixes #261

### DIFF
--- a/op-acceptor/config.go
+++ b/op-acceptor/config.go
@@ -20,6 +20,7 @@ type Config struct {
 	RunInterval     time.Duration // Interval between test runs
 	RunOnce         bool          // Indicates if the service should exit after one test run
 	AllowSkips      bool          // Allow tests to be skipped instead of failing when preconditions are not met
+	DefaultTimeout  time.Duration // Default timeout for individual tests, can be overridden by test config
 
 	Log log.Logger
 }
@@ -53,6 +54,8 @@ func NewConfig(ctx *cli.Context, log log.Logger, testDir string, validatorConfig
 	runInterval := ctx.Duration(flags.RunInterval.Name)
 	runOnce := runInterval == 0
 
+	defaultTimeout := ctx.Duration(flags.DefaultTimeout.Name)
+
 	return &Config{
 		TestDir:         absTestDir,
 		ValidatorConfig: absValidatorConfig,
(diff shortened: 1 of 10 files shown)


op-acceptor/config.go
```diff
@@ -20,6 +20,7 @@ type Config struct {
 	RunInterval     time.Duration // Interval between test runs
 	RunOnce         bool          // Indicates if the service should exit after one test run
 	AllowSkips      bool          // Allow tests to be skipped instead of failing when preconditions are not met
+	DefaultTimeout  time.Duration // Default timeout for individual tests, can be overriden by test config
```
janjakubnanista: Tiniest of things - why not just Timeout?

op-acceptor/flags/flags.go
```diff
@@ -53,6 +54,12 @@ var (
 		EnvVars: opservice.PrefixEnvVar(EnvVarPrefix, "ALLOW_SKIPS"),
 		Usage:   "Allow tests to be skipped instead of failing when preconditions are not met.",
 	}
+	DefaultTimeout = &cli.DurationFlag{
```
janjakubnanista, line 57: The same from above - the sorta standard across all frameworks is --timeout, i'd just name it that

scharissis (changes requested):
Awesome, thank you!

A few notes/questions:
1. How would one override the timeout of an individual test?
2. Can you please add some tests showing this all works?
3. CI is failing due to lint errors. (To check them locally run `golangci-lint run -E goimports,sqlclosecheck,bodyclose,asciicheck,misspell,errorlint --timeout 5m -e "errors.As" -e "errors.Is" ./...`)

op-acceptor/config.go
scharissis: Just adding my opinion to this: if we can override it per-test/suite/gate then I like having 'Default' in the name. Because if you ran it with `op-acceptor --timeout 5m` and then a test timed out after 1m (because it has an override), that'd be confusing. Whereas if it were labelled `--default-timeout` then that would explain things.

teddyknox: @scharissis explained well

op-acceptor/flags/flags.go
teddyknox: @scharissis covered well my thoughts https://github.com/ethereum-optimism/infra/pull/262#discussion_r2011030586

teddyknox:
Thanks for your comments @scharissis, re-requesting!

scharissis (approved):

PR #263, opened 2025-03-27
fix(op-signer): allow anon auth when mTLS is disabled (for local testing only)

ddaws:
**Description**

This MR adds `func NewAnonMiddleware() oprpc.Middleware` which sets the `ClientInfo.ClientName` to `anonymous` when mTLS auth is disabled. This is needed to allow disabling TLS auth locally. 

**Tests**

You can test by creating a config.yaml like 

```yaml
provider: LOCAL
auth:
  - name: anonymous          # The key needs to be explicitly allowed for anonymous usage
    key: tls/ec_private.pem
```

Then post to op-signer to see the signing results

```bash
$ curl -X POST -H "Content-Type: application/json" -d @test-rpc.json http://localhost:8080
[
  {
    "jsonrpc": "2.0",
    "id": "1",
    "result": "0x02f89d82053980010182753094000000000000000000000000000000000000aaaa0180f838f794000000000000000000000000000000000000aaaae1a0000000000000000000000000000000000000000000000000000000000000000001a072b0f296886fe5eeb4b18cba71236063dfe520789d8fd77fa3a8ede175ba5c68a006a19fe680bda685effac26c18e661d7fd0f578fafb14c24b1e72ff80dc7cd45"
  },
  {
    "jsonrpc": "2.0",
    "id": "1",
    "error": {
      "code": -32010,
      "message": "nonce not specified"
    }
  }
]

```

### DIFF
--- a/op-signer/app.go
+++ b/op-signer/app.go
@@ -147,6 +147,7 @@ func (s *SignerApp) initRPC(cfg *Config) error {
 		authMiddleware = service.NewAuthMiddleware()
 	} else {
 		s.log.Warn("TLS disabled. This disables authentication and is INSECURE! Please enable TLS in production environments!")
+		s.log.Info("Using anonymous authentication. You must explicitly set the 'auth[].name' config field to 'anonymous' to allow unauthenticated anonymous usage.")
 		authMiddleware = service.NewAnonMiddleware()
 	}
 
--- a/op-signer/app.go
+++ b/op-signer/app.go
@@ -113,6 +113,7 @@ func (s *SignerApp) initMetrics(cfg *Config) error {
 
 func (s *SignerApp) initRPC(cfg *Config) error {
 	var httpOptions = []httputil.Option{}
+	var authMiddleware oprpc.Middleware
 
 	if cfg.TLSConfig.Enabled {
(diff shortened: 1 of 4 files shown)


edobry:
LGTM overall, tested & confirmed working.

however, what should happen if the config file's `auths` list has multiple `anonymous` entries? i tested this and it seemed to just pick one of the entries arbitrarily, should there be some validation here?

additionally, seems like its possible to add `auths` entries with a non-`anonymous` name when TLS is disabled, which will never get used; similarly, should there be validation around this?

ddaws:
> however, what should happen if the config file's auths list has multiple anonymous entries? i tested this and it seemed to just pick one of the entries arbitrarily, should there be some validation here?

I think this is fine and there doesn't need to be validation added here. This would be the same if there were duplicate hostname entries, so I would expect similar behavior. 

> additionally, seems like its possible to add auths entries with a non-anonymous name when TLS is disabled, which will never get used; similarly, should there be validation around this?

I don't think we need to validate for this. I think this is fine, and they just wont be usable, and probably shouldn't be usable. 

I'm happy to add some validation in another PR. I think this should just result in warnings being logged

PR #265, opened 2025-03-28
feat(op-signer/gen-local-creds): parameterize gen script

edobry:
This PR parameterizes and renames the `gen-local-tls.sh` script to `gen-local-creds.sh` to enable users to generate a subset of credentials, intended to enable its usage in [`optimism-package`](https://github.com/ethpandaops/optimism-package/pull/207). It adds a CLI parameter taking one of `{ca,client_tls,client_signing_key,all}`, determining what the script will do. It also enables generating TLS credentials for multiple client hostnames in a single run, which is useful for Kurtosis.

In support of https://github.com/ethereum-optimism/platforms-team/issues/581.


op-signer/gen-local-creds.sh
```diff
@@ -0,0 +1,189 @@
...
+    CLIENT_HOSTNAMES="$*"
+    if [ -z "$CLIENT_HOSTNAMES" ]; then
+        CLIENT_HOSTNAMES="localhost"
+    fi
+    printf "\nProcessing clients: %s\n" "$CLIENT_HOSTNAMES"
+}
+
+process_clients() {
+    generator="$1"
+    for hostname in $CLIENT_HOSTNAMES; do
```
ddaws, line 134: It looks like this will generate a certificate per hostname in the case of `process_clients  generate_client_tls`. Why create a cert per hostname instead of including multiple DNS SAN in the single cert? I think multiple SAN is the real world way a client would be identified by multiple DNS names

op-signer/gen-local-creds.sh
```diff
@@ -0,0 +1,189 @@
...
+        -CA "$CA_CERT" \
+        -CAkey "$CA_KEY" \
+        -CAcreateserial \
+        -out "$clientDir/$CLIENT_TLS_CERT" \
+        -days 3 \
+        -extensions san \
+        -extfile "$confFile"
+}
+
+generate_client_key() {
```
ddaws: Can we rename this to 

```suggestion
generate_signing_key() {
```

The secp256k1 isn't actually the client's key, it's the key that will be maintained by op-signer. In the wild this key should never be accessible to the client and op-signer will provide crypto-as-a-service to the client

op-signer/gen-local-creds.sh
```diff
@@ -0,0 +1,189 @@
...
+
+case "$TARGET" in
+    "ca")
+        generate_ca true
+        ;;
+    "client_tls")
+        setup_client_hostnames "$@"
+        process_clients generate_client_tls
+        ;;
+    "client_key")
```
ddaws: Same here. How about 

```suggestion
    "signing_key")
```

ddaws:
@edobry have you thought about moving this to the Makefile? I think this is a good candidate because the command generates real targets (files). `make build` could depend on the keys being generated

op-signer/gen-local-creds.sh
```diff
@@ -0,0 +1,189 @@
...
+            echo "Error: OpenSSL is not installed locally. Please install OpenSSL or use Docker by setting OP_SIGNER_GEN_TLS_DOCKER=true"
+            exit 1
+        fi
+        openssl "$@"
+    fi
+}
+
+generate_ca() {
+    local force="$1"
+    [ "$force" = "true" ] || [ ! -f "$CA_CERT" ] || return 0
```
ddaws, line 49: Make does this by default, which is why I was thinking it might make sense to move this to the Makefile

ddaws (changes requested):

op-signer/gen-local-creds.sh
edobry: the intention is to create multiple keys to be used by multiple services (see [here](https://github.com/ethpandaops/optimism-package/pull/207/files#diff-8952ca36c0cbca2168c99c185e468d02357daaab464a93cd7bf58bf8f3b90eb0R338))

op-signer/gen-local-creds.sh
edobry: it would probably be good to move to the Makefile and leverage this feature, however I'm loathe to invest in it as most OP Labs projects have transitioned to using [just](https://github.com/casey/just). could be good to migrate this one as well in a future PR

scharissis (approved):

PR #279, opened 2025-04-10
explicitly handle some backend responses so we don't mark as unservic…

Melvillian:
…eable

**Description**

I added handling for additional error types from our backends so that we can reduce the number of unserviceable requests.

**Tests**

I couldn't find any existing tests for this functionality, and it's not adding new functionality (but rather constraining existing functionality)

**Additional context**

After some closer looks at how we handle unserviceable requests in proxyd, and the error logs that involve unserviceable requests, I think handling these 2 types of errors will eliminate 2 of the 3 main categories of unserviceable requests we (Unichain) see. The 3rd category involves problems with how we deploy and swap out old backends with new backends, and we will not handle that PR here.

### DIFF
--- a/proxyd/backend.go
+++ b/proxyd/backend.go
@@ -1428,6 +1428,11 @@ func (bg *BackendGroup) ForwardRequestToBackendGroup(
 
 			res, err = back.Forward(ctx, rpcReqs, isBatch)
 
+			// below are errors that we explicitly handle so that we don't
+			// want to mark this request as unserviceable (unserviceable
+			// requests indicate a problem with our backends, but there is
+			// nothing wrong with our backends in this case)
+
 			if errors.Is(err, ErrConsensusGetReceiptsCantBeBatched) ||
 				errors.Is(err, ErrConsensusGetReceiptsInvalidTarget) ||
 				errors.Is(err, ErrMethodNotWhitelisted) {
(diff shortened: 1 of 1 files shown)


proxyd/backend.go
```diff
@@ -1444,6 +1449,35 @@ func (bg *BackendGroup) ForwardRequestToBackendGroup(
 					error:    err,
 				}
 			}
+			// we check for "request body too large" when first serving a request,
+			// so this is a special case where the backend has its own rules around
+			// request body size and returns a 413 error. We've seen this with quicknode
+			if errors.Is(err, ErrRequestBodyTooLarge) {
+				return &BackendGroupRPCResponse{
+					RPCRes:   nil,
+					ServedBy: servedBy,
```
jelias2, line 1458: Any particular reason that servedBy is left in this backend group response?

jelias2 (commented):
Overall, LGTM. 

As maintainers we may follow up this PR and group all of the errors together in the conditional on 1436 as long as servedBy isn't explicitly required.

Let us know your thoughts

Thanks for the contribution!

proxyd/backend.go
Melvillian: My thinking was that because this was actually served by a backend, and the error is specific to that backend (i.e. quicknode), that it should have a non-nil`ServedBy`.

I thought the `ErrBackendResponseTooLarge` and `ErrMethodNotWhitelisted` if blocks should also return a `ServedBy`, but I didn't write that code so I didn't change it.

I don't have strong thoughts on this, though. So if there is some existing reason why those other if cases leave `ServedBy` as `nil`, I'm happy to this `ServedBy` to be `nil`. Actually in that case I will group them all under the single `if` check in order to make the code more concise. But I would like to leave in the `comments` above each error type, so it's clear to readers why we're explicitly handling them.

jelias2: I like your idea about adding served by to ErrBackendResponseTooLarge, ErrMethodNotWhitelisted, ErrRequestBodyTooLarge to be returned since these are backend specific. 

Can you group those errors together in a conditional and the errors which should not contain a servedBy together in another conditional?

Melvillian: done

jelias2 (approved):

PR #288, opened 2025-04-18
chore: Update `go-ethereum` dep in proxyd

clabby:
## Overview

Updates the `go-ethereum` dependency in `proxyd` to the latest version (`v1.15.8`). [`v1.15.1`](https://github.com/ethereum/go-ethereum/releases/tag/v1.15.1) included support for EIP-7702 transactions in the transaction pool, and we're currently still one release behind.

This is currently causing an issue on Sepolia chains that have activated Isthmus, where EIP-7702 transactions aren't being accepted.


clabby:
Heads up, btw - new version of geth includes https://github.com/ethereum/go-ethereum/pull/31434, and the default signer now disables legacy transactions. Updated one of the tests to account for this new behavior.

proxyd/integration_tests/sender_rate_limit_test.go
```diff
@@ -57,8 +57,16 @@ func TestSenderRateLimitValidation(t *testing.T) {
...
 		t.Run(name, func(t *testing.T) {
 			res, _, err := client.SendRequest([]byte(body))
-			require.NoError(t, err)
-			RequireEqualJSON(t, []byte(expResponseBody), res)
+
+			// An update to geth now causes this case to fail.
+			// Legacy transactions are now disabled by default on the modern signer: https://github.com/ethereum/go-ethereum/pull/31434
+			if strings.Contains(name, "no chain id (pre eip-155)") {
+				require.Error(t, err)
+				return
```
protolambda, line 65: We should manually enable legacy txs again. Otherwise the deterministic deployment txs that people use, which don't have a chainID for cross-chain compat, won't be possible to confirm anymore.

protolambda (changes requested):
Change LGTM, but we can't just stop accepting legacy transactions, because they still have an important use-case for preinstall-like contracts that are deployed with the deterministic-legacy-tx method instead of a factory on every chain.

PR #295, opened 2025-04-22
feat(op-acceptor): add network name to metrics

### LINKED ISSUE #193: [nat] metrics - add network dimension
When we emit metrics we don't know the 'network' we're running against. Let's either find a way to obtain it and add it to our metrics, or remove it from our metrics entirely.

eg. _runner.go_:
```
func (r *runner) RunTest(metadata types.ValidatorMetadata) (*TestResult, error) {
	...
	metrics.RecordValidation("todo", r.runID, metadata.ID, metadata.Type, result.Status) // TODO: handle network
        ...
}
```

scharissis:
**Description**

- Parses the network name and adds it to our metrics
- Removes simple-devnet.json (not needed and unrelated to this change)

**Metadata**

- Fixes #193


op-acceptor/nat.go
```diff
@@ -92,9 +103,53 @@ func New(ctx context.Context, config *Config, version string, shutdownCallback f
...
+//   - If the file exists, it tries to parse it as JSON and extract the "name" field
+//   - If parsing fails or the file doesn't exist, it falls back to using the filename
+//     with both the file extension and "-devnet" suffix removed if present
+func extractNetworkName(envURL string) string {
+	if envURL == "" {
+		return "unknown" // Default if not set
+	}
+
+	// Handle Kurtosis URLs like "kt://isthmus-devnet"
+	if strings.HasPrefix(envURL, "kt://") {
```
sigma: The URL handling should be a bit more involved, to reflect the options at https://github.com/ethereum-optimism/optimism/blob/develop/devnet-sdk/shell/env/devnet.go#L45

ktnative should be supported, but also the file path is just a special case of the more generic file:// URL (file  being the default protocol).
So I think you should always parse the URL, then adjust the logic based on the parsing result.

sigma (changes requested):

op-acceptor/nat.go
scharissis: Thanks. As discussed outside of Github, I'll swap to use devnet-sdk's LoadEnvFromURL as this is getting more complex than I'd like. :)

bitwiseguy (approved):

sigma (approved):

PR #300, opened 2025-04-29
feat: interop message validation

### LINKED ISSUE #358: proxyd: (interop message validation) Interpretability of interop transactions and supervisor-backed validation against them
A/C:
- Proxyd should be able to interpret incoming interop transaction
- It should be supervisor-aware to validate that incoming transaction

yashvardhan-kukreja:

**Description**

**Tests**

**Additional context**

**Metadata**

- Fixes https://github.com/ethereum-optimism/infra/issues/358


proxyd/server.go
```diff
@@ -388,6 +394,84 @@ func (s *Server) HandleRPC(w http.ResponseWriter, r *http.Request) {
...
+		go func(b miner.BackendWithInterop) {
+			defer wg.Done()
+			err := b.CheckAccessList(ctx, interopAccessList, interoptypes.CrossUnsafe, interoptypes.ExecutingDescriptor{
+				Timestamp: uint64(time.Now().Second() + 1000),
+			})
+			resultChan <- err
+		}(validatingBackend)
+	}
+
+	// goroutine which waits for all the sender goroutines created above to be done, and drain and close the resultChan
```
jelias2: Why use a multi-call strategy instead of just a normal iterative approach? 

Was this behavior explictly requested by the protocol team?

proxyd/server.go
```diff
@@ -388,6 +394,84 @@ func (s *Server) HandleRPC(w http.ResponseWriter, r *http.Request) {
...
+	resultChan := make(chan error, len(s.interopValidatingBackends))
+
+	// concurrently broadcast the checkAccessList operation to all the validating backends
+	var wg sync.WaitGroup
+	for _, validatingBackend := range s.interopValidatingBackends {
+		wg.Add(1)
+		go func(b miner.BackendWithInterop) {
+			defer wg.Done()
+			err := b.CheckAccessList(ctx, interopAccessList, interoptypes.CrossUnsafe, interoptypes.ExecutingDescriptor{
+				Timestamp: uint64(time.Now().Second() + 1000),
```
jelias2: Whats the timestamp in this body, should the 1000ms be configurable?

proxyd/server.go
```diff
@@ -388,6 +394,84 @@ func (s *Server) HandleRPC(w http.ResponseWriter, r *http.Request) {
...
+			})
+			resultChan <- err
+		}(validatingBackend)
+	}
+
+	// goroutine which waits for all the sender goroutines created above to be done, and drain and close the resultChan
+	go func() {
```
jelias2: Please add some prom metrics for these types of RPC calls for the b.CheckAccessList calls

proxyd/server.go
jelias2: If one supervisor is down in this case, then that could fail all interop transactions since it would return errors for all?

proxyd/server.go
jelias2: Also a debug log for the final output of the function would be nice

proxyd/server.go
```diff
@@ -388,6 +394,84 @@ func (s *Server) HandleRPC(w http.ResponseWriter, r *http.Request) {
...
+				Code:    rpcErr.Error.Code,
+				Message: rpcErr.Error.Message,
+				Data:    rpcErr.Error.Data,
+
+				HTTPErrorCode: httpErr.StatusCode,
+			}
+		}
+
+		return &RPCErr{
+			Code:          -32602,
```
jelias2: Can the error code be documented as a constant and named to reflect the type of error

ex. const RPCErrorInvalidInterop := -32602

proxyd/server.go
yashvardhan-kukreja: After confirming from Axel: it can be any timestamp slightly in the future.

Behind the scenes: it backs the access-list time-check which supervisor performs
- It must be higher than the timestamp associated with the access-list entries
- Must be lower than the expiry of the message (expiry being a pretty high value)

proxyd/server.go
jelias2: A nice to have improvement could be health checking the supervisors, and then using the filtered list to send the checks. 

Out of scope for now, just thinking outloud

yashvardhan-kukreja: Just HA of supervisors and slightly better IO performance. Wasn't so hard to implement and test, so didn't hesitate to skip it in the name of PMO.

proxyd/integration_tests/interop_validation_test.go
```diff
@@ -0,0 +1,137 @@
...
+		require.Equal(t, 200, code1, "iteration %d: response observed: %s", i, string(res1))
+		RequireEqualJSON(t, []byte(dummyHealthyRes), res1)
+	}
+}
+
+func TestInteropValidationBadPath(t *testing.T) {
+	goodBackend := NewMockBackend(SingleResponseHandler(200, dummyHealthyRes))
+	defer goodBackend.Close()
+
+	require.NoError(t, os.Setenv("GOOD_BACKEND_RPC_URL", goodBackend.URL()))
```
jelias2, line 66: Is this good backend still required in the Bad Path test?

proxyd/server.go
jelias2: Can you double check with protocol folks that multicall is the desired behavior

yashvardhan-kukreja: Sure, sent the message. Will update upon the response.

jelias2 (approved):

PR #310, opened 2025-05-08
feat(op-acceptor): integrate devnet-sdk telemetry capabilities

sigma:
**Description**

This change integrates devnet-sdk telemetry capability, so we can gain
visibility over the tests through the lens of op-acceptor.

For the most part, it consists of simply passing a context around
where it wasn't before, and to create telemetry spans in the right
places to reflect the overall workflow.

And of course, making sure that the `go test` invocations that
ultimately run the tests are provided with the right context to
ensure continuity of tracing.

**Additional context**

Result looks like this:
Screenshot 2025-05-06 at 19 56 43

**Metadata**

- Part of https://github.com/ethereum-optimism/platforms-team/issues/613


op-acceptor/runner/runner.go
```diff
@@ -212,7 +214,11 @@ func (r *runner) groupValidatorsByGate() map[string][]types.ValidatorMetadata {
 }
 
 // processGate handles the execution of a single gate and its tests
-func (r *runner) processGate(gateName string, validators []types.ValidatorMetadata, result *RunnerResult) error {
+func (r *runner) processGate(ctx context.Context, gateName string, validators []types.ValidatorMetadata, result *RunnerResult) error {
+	tracer := otel.Tracer("runner")
```
janjakubnanista: `op-acceptor/runner` maybe?

sigma: I mean maybe, but:
- the tracer name is already nested under the service name (which is "op-acceptor")
- it's seldom used anyway

That said, I definitely shouldn't be creating multiple tracers with the same name. Will fix that

janjakubnanista: Aaah okay so it's nested

scharissis (approved):
Amazing, thanks!

PR #336, opened 2025-05-12
feat(op-acceptor): optionally embed faucet

sigma:
**Description**

This change detects if the provided devnet has a faucet deployed.
If not:
- it embeds its own, using op-faucet as a library
- it mutates the devnet environment to augment it with the faucet information
- it points the running tests to the mutated environment

The embedded faucet is backed by a "l2Faucet" key, provided by the devnet.

We also introduce a slightly more generic notion of op-acceptor
"addons", to expose arbitrary augmentations to devnets test-side.

**Tests**

Manual validation using:
- a devnet *without* a faucet
- a test requiring one


scharissis (approved):
LGTM, ty!

PR #337, opened 2025-05-13
feat(proxyd): req params size limit

### LINKED ISSUE #361: proxyd: (interop message validation) Bounding/limiting the requests’ size housing the interop params
Many a times, there's a direct correlation between the size of the incoming transaction req body and the payload which gets sent to the supervisor for validation.

And currently, proxyd can potentially exhaust the supervisor as well as itself with some static checks, if an arbitrarily large incoming transaction req is processed blindly.

A/C:
- In favor of not exhausting the backing supervisor, proxyd should have an interop specific check on the size of the incoming transaction as early in the message-validation path as possible.

yashvardhan-kukreja:

**Description**

**Tests**

Included integration tests

**Additional context**

**Metadata**

- Fixes https://github.com/ethereum-optimism/infra/issues/361

### DIFF
--- a/proxyd/server.go
+++ b/proxyd/server.go
@@ -464,7 +464,7 @@ func (s *Server) validateInteropSendRpcRequest(ctx context.Context, rpcReq *RPCR
 	}
 
 	if s.interopValidatingConfig.AccessListSizeLimit > 0 && len(interopAccessList) > s.interopValidatingConfig.AccessListSizeLimit {
-		log.Info(
+		log.Error(
 			"interop access list exceeds maximum size limit",
 			"size", len(interopAccessList),
 			"max_size", s.interopValidatingConfig.AccessListSizeLimit,
--- a/proxyd/server.go
+++ b/proxyd/server.go
@@ -51,7 +51,7 @@ const (
 	defaultMaxUpstreamBatchSize      = 10
 	defaultRateLimitHeader           = "X-Forwarded-For"
 	defaultInteropValidationStrategy = FirstSupervisorStrategy
-	defaultInteropReqParamsSizeLimit = 128 * 1024 // 128KB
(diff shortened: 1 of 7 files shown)


proxyd/server.go
```diff
@@ -395,6 +397,30 @@ func (s *Server) HandleRPC(w http.ResponseWriter, r *http.Request) {
 	writeRPCRes(ctx, w, backendRes[0])
 }
 
+func senderReqSizeLimitCheck(ctx context.Context, rpcReq *RPCReq, maxTxSize int) error {
```
jelias2: This is a bit confusing to me. The config is `req_params_size_limit` but the code is calculating len of parameters not the actual size?

In a different part of the code it looks like to calculate the maxbody size we've use [io.ReadAll](https://github.com/ethereum-optimism/infra/blob/6d9a62904967ce43cfa573687832962c26486cdd/proxyd/server.go?plain=1#L314), should that pattern be used again?

yashvardhan-kukreja: Idm, just went ahead with the existing one because that's the only dynamic thing in the request which gets parsed to be eventually forwarded to the supervisor.

And the 2 hex char -> 1 byte correspondence seemed like a decent ground for using `len(params)` to base the limit on.

Feel free to let me know your thoughts, I am not strongly inclined with any approach.

proxyd/server.go
```diff
@@ -395,6 +397,32 @@ func (s *Server) HandleRPC(w http.ResponseWriter, r *http.Request) {
 	writeRPCRes(ctx, w, backendRes[0])
 }
 
+// reqSizeLimitCheck is a function which helps define, check and limit the size of the incoming request beyond the "max_request_body_size_bytes" setting.
```
jelias2, line 401: <3

jelias2 (approved):

PR #340, opened 2025-05-13
feat(proxyd): interop access list inbox entries client-side deduplication

### LINKED ISSUE #357: proxyd: (interop-message-validation) Possible optimizations in proxyd to avoid/soften the IO to supervisor
There's a lot of preflight processing which supervisor does on its end to before even validating what's backed by the supervisor DB.

Those static preprocessing routines can be potentially performed in proxyd to reject the incoming entries failing them without needing to IO with the supervisor.

Moreover, the inbox entries decided to be sent to supervisor for validation can be potentially duplicated, which unnecessarily exhausts the supervisor further.

A/C:
- proxyd to perform static access parsing checks on its end
- proxyd to perform deduplication of inbox entries on a per-access basis before sending them to supervisor.

yashvardhan-kukreja:

**Description**

The following is the old description. Please look at ethereum-optimism/optimism#349 which includes the description of the updated logic.

-----

This PR introduces the capability of performing an ordered deduplication of inbox entries within the access list of an interop transaction.

Context

> Interop transactions are simple sendRawTransaction with one or more AccessList = CrossL2InboxAddress
> Each AccessList further points to multiple StorageHash or InboxEntries.
> As the request arrives at the proxyd, all the AccessList corresponding to CrossL2InboxAddress are parsed and all of their respective StorageHash/InboxEntries are flattened and joined into one list of inboxEntries.
> This list is later sent to the supervisor to evaluate where supervisor evaluates them individually (kind of).

An area of  optimization here is to deduplicate all the inbox entries before sending to the supervisor.

For example, if the incoming payload is

```
method: sendRawTransaction
accessList: [
       {
               Address: CrossL2Inbox,
               StorageHashes: ["a", "b", "a", "a", "b"]
       },
       {
               Address: CrossL2Inbox,
               StorageHashes: ["b", "b"]
       }
]
```

Without deduplication, the following list of inbox entries would go ["a", "b", "a", "a", "b", "b", "b"]

After deduplication, the following list of inbox entries would propagate ["a", "b"]

The order is preserved here for the sake of how inbox entries' order is used on supervisor's end.

**Tests**

**Additional context**

**Metadata**

- Fixes https://github.com/ethereum-optimism/infra/issues/357

### DIFF
--- a/proxyd/config.go
+++ b/proxyd/config.go
@@ -233,6 +233,7 @@ type InteropValidationConfig struct {
 	Strategy            InteropValidationStrategy `toml:"strategy"`
 	ReqSizeLimit        int                       `toml:"req_size_limit"`
 	AccessListSizeLimit int                       `toml:"access_list_size_limit"`
+	RateLimit           SenderRateLimitConfig     `toml:"sender_rate_limit"`
 }
 
 type InteropValidationStrategy string
--- a/proxyd/integration_tests/interop_validation_test.go
+++ b/proxyd/integration_tests/interop_validation_test.go
@@ -6,6 +6,7 @@ import (
 	"math/big"
 	"os"
 	"testing"
+	"time"
 
 	"github.com/ethereum-optimism/infra/proxyd"
 	"github.com/ethereum-optimism/optimism/op-service/eth"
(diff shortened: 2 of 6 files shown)


jelias2 (approved):
If metrics are appliacable, please them it will be important so we can judge how well interop tx filtering is wokring

jelias2 (approved):

axelKingsley (approved):
Reviewed the deduplication logic specifically, which is sound -- the validation for [a a] is the same as the validation for [a], so message size can be kept down which is great. The code which deduplicates is sound as well.

PR #343, opened 2025-05-13
feat(proxyd): interop message validation sender rate limit

### LINKED ISSUE #360: proxyd: (interop-message-validation) Interop-specific sender rate limits scoped per client
Although proxyd has support for global and even method-specific override rate limits, there still be different traffic patterns between a non-interop sendRawTransaction v/s an interop one.

Therefore, proxyd should have a way to define rate limits scoped per client dedicated to interop message validation

A/C:
- Proxyd has support for configuring and exercising the interop-specific rate limits on the caller.

yashvardhan-kukreja:

**Description**

**Tests**

**Additional context**

**Metadata**

- Fixes https://github.com/ethereum-optimism/infra/issues/360

### DIFF
--- a/proxyd/server.go
+++ b/proxyd/server.go
@@ -74,6 +74,7 @@ type Server struct {
 	mainLim                 FrontendRateLimiter
 	overrideLims            map[string]FrontendRateLimiter
 	senderLim               FrontendRateLimiter
+	interopSenderLim        FrontendRateLimiter
 	allowedChainIds         []*big.Int
 	limExemptOrigins        []*regexp.Regexp
 	limExemptUserAgents     []*regexp.Regexp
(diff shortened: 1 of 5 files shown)


proxyd/server.go
```diff
@@ -915,7 +931,7 @@ func convertSendReqToSendTx(ctx context.Context, req *RPCReq) (*types.Transactio
 	return tx, nil
 }
 
-func (s *Server) rateLimitSender(ctx context.Context, req *RPCReq) error {
+func (s *Server) genericRateLimitSender(ctx context.Context, req *RPCReq, lim FrontendRateLimiter) error {
```
jelias2, line 938: Doesn't this need a pointer type for the FrontendRateLimitSender? Otherwise the struct will be copied and not actually updated?

yashvardhan-kukreja: FrontendRateLimiter is an interface.

All the inherent implementations of it are pointer-based only, so we end up indirectly passing around the pointer to the backing rate limiter.

(And I can't imagine for any newer in-the-future rate limiter to not be pointer-based unless its backed by a remote backend.)

jelias2 (approved):

PR #349, opened 2025-05-14
fix: interop deduplication of inbox entries of a per-access basis

yashvardhan-kukreja:

**Description**

Supervisor doesn't process inbox entries on a one-by-one entry basis.

Instead, it looks at a two-to-three inbox entries at a time, wraps them into one "Access" object, and processes them overall on a per-Access object basis.

This PR fixes the deduplication logic to respect that.

in simple words,

Before this PR

["0x01", "0x02", "0x03", "0x01", "0x03", "0x01", "0x03"] inbox entries (hashes of access lists) would be wrongfully deduplicated to ["0x01", "0x02", "0x03"]

After this PR

["0x01", "0x02", "0x03", "0x01", "0x03", "0x01", "0x03"] inbox entries would be rightfully deduplicated to ["0x01", "0x02", "0x03", "0x01", "0x03"]

This is because the sub list ["0x01", "0x02", "0x03"] and ["0x01", "0x03"] which clearly aren't duplicates.

**Tests**

**Additional context**

**Metadata**

- Fixes #[Link to Issue]

### DIFF
--- a/proxyd/integration_tests/interop_validation_test.go
+++ b/proxyd/integration_tests/interop_validation_test.go
@@ -358,53 +358,128 @@ func TestInteropValidation_Deduplication(t *testing.T) {
 
 	require.NoError(t, os.Setenv("GOOD_BACKEND_RPC_URL", goodBackend.URL()))
 
-	config := ReadConfig("interop_validation")
-	config.SenderRateLimit.Limit = math.MaxInt // Don't perform basic rate limiting in this test since we're only testing interop validation.
-
-	config.InteropValidationConfig.AccessListSizeLimit = 2 // only 2 entries are allowed in the access list
-
 	validatingBackend1 := NewMockBackend(SingleResponseHandler(200, dummyHealthyRes))
 	defer validatingBackend1.Close()
 
 	validatingBackend2 := NewMockBackend(SingleResponseHandler(200, dummyHealthyRes))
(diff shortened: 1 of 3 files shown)


yashvardhan-kukreja:
The first commit brings in enhancements to the test catching the edge-case which fails the current implementation with the following logs

```
=== RUN   TestInteropValidation_Deduplication
    /Users/ykukreja/OPLabs/Projects/infra/proxyd/integration_tests/interop_validation_test.go:464:
                Error Trace:    /Users/ykukreja/OPLabs/Projects/infra/proxyd/integration_tests/interop_validation_test.go:464
                Error:          Not equal:
                                expected: 413
                                actual  : 200
                Test:           TestInteropValidation_Deduplication
                Messages:       the request should have failed because of the expectation of the deduplicated entries being 5
--- FAIL: TestInteropValidation_Deduplication (0.01s)
FAIL
FAIL    github.com/ethereum-optimism/infra/proxyd/integration_tests     0.625s
```

yashvardhan-kukreja:
The second commit brings in the actual fix which does end up passing the test which were failing against the previous approach.

```
=== RUN   TestInteropValidation_Deduplication
--- PASS: TestInteropValidation_Deduplication (0.03s)
PASS
ok      github.com/ethereum-optimism/infra/proxyd/integration_tests     0.651s
```

PR #352, opened 2025-05-14
fix(op-acceptor): add build dependencies and increase logs

### LINKED ISSUE #374: op-acceptor: Output logs to the console and enable extra logs for tests when running in CI
**Is your feature request related to a problem? Please describe.**
We currently need to SSH into the container in order to see the test logs. This means that debugging the test logs for op-acceptor is very cumbersome on CI as we can't easily figure out what happened.
Additionally we are only able to see the `info` level of the test logs which means we may be missing critical information.

**Describe the solution you'd like**
We should enable op-acceptor to output the logs to the console on CI and additionally enable custom log levels for the tests.

**Describe alternatives you've considered**
SSHing into the container, though it may not be straighforward for everybody when dealing with multiple runs.

serpixel:
**Description**

- Add git and other build dependencies to the op-acceptor docker image. 
- Increase the timeout for listing tests as it downloads dependencies and increase logs
- Add extra logging for tests

**Tests**

Added a test to verify the paths checks logic.

**Additional context**

When running the tests sometimes we may want to get extra information on these. Right now we need to ssh into the container, but this PR adds a flag to dump the logs into the console.

Requires https://github.com/ethereum-optimism/optimism/pull/16050 for the test log level.

**Metadata**

- Fixes #[Link to Issue]

### DIFF
--- a/op-acceptor/Dockerfile
+++ b/op-acceptor/Dockerfile
@@ -14,7 +14,7 @@ RUN just build
 
 FROM golang:1.23.5-alpine3.21
 
-RUN apk add --no-cache ca-certificates build-base
+RUN apk add --no-cache ca-certificates build-base git
 
 # Increase the proxy connect timeout to 60 seconds
 ENV GOPROXYCONNECTTIMEOUT=60s
@@ -27,6 +27,7 @@ ENV GOPATH="/go"
 ENV PATH="${GOPATH}/bin:${PATH}"
 
 WORKDIR /app
+
 RUN addgroup -S app && adduser -S app -G app && \
     mkdir -p /devnets && \
     chown -R app:app /devnets
--- a/op-acceptor/Dockerfile
+++ b/op-acceptor/Dockerfile
@@ -14,6 +14,8 @@ RUN just build
 
 FROM alpine:3.21
 
+RUN apk add --no-cache ca-certificates
+
 # Increase the proxy connect timeout to 60 seconds
 ENV GOPROXYCONNECTTIMEOUT=60s
 
--- a/op-acceptor/Dockerfile
+++ b/op-acceptor/Dockerfile
@@ -1,5 +1,8 @@
 FROM golang:1.23.5-alpine3.21 AS builder
 
+# Increase the proxy connect timeout to 60 seconds
+ENV GOPROXYCONNECTTIMEOUT=60s
+
 WORKDIR /app
 
 # Install dependencies first for better caching
@@ -10,6 +13,9 @@ COPY op-acceptor/ .
 RUN just build
 
 FROM alpine:3.21
+
+# Increase the proxy connect timeout to 60 seconds
+ENV GOPROXYCONNECTTIMEOUT=60s
 
 # Install Go binary
 # (we copy Go directly from the builder stage for simplicity and consistency)
@@ -30,7 +36,12 @@ chown -R app:app /go
 
 COPY --from=builder /app/bin/op-acceptor /app/
 COPY op-acceptor/example-validators.yaml /app/
+
+# Set ownership of the /app directory to allow the application to write logs and other files at runtime
+RUN chown -R app:app /app/
+
 USER app
 
 VOLUME /devnets
+
 ENTRYPOINT ["/app/op-acceptor"]
--- a/op-acceptor/Dockerfile
+++ b/op-acceptor/Dockerfile
@@ -1,8 +1,5 @@
 FROM golang:1.23.5-alpine3.21 AS builder
 
-# Increase the proxy connect timeout to 60 seconds
-ENV GOPROXYCONNECTTIMEOUT=60s
-
 WORKDIR /app
 
 # Install dependencies first for better caching
@@ -15,9 +12,6 @@ RUN just build
 FROM golang:1.23.5-alpine3.21
 
 RUN apk add --no-cache ca-certificates build-base git
-
-# Increase the proxy connect timeout to 60 seconds
-ENV GOPROXYCONNECTTIMEOUT=60s
 
 # Install Go binary
 # (we copy Go directly from the builder stage for simplicity and consistency)
--- a/op-acceptor/Dockerfile
+++ b/op-acceptor/Dockerfile
@@ -12,9 +12,9 @@ COPY op-acceptor/ .
 
 RUN just build
 
-FROM alpine:3.21
+FROM golang:1.23.5-alpine3.21
 
-RUN apk add --no-cache ca-certificates
+RUN apk add --no-cache ca-certificates build-base
 
 # Increase the proxy connect timeout to 60 seconds
 ENV GOPROXYCONNECTTIMEOUT=60s
--- a/op-acceptor/runner/logging_test.go
+++ b/op-acceptor/runner/logging_test.go
@@ -205,8 +205,8 @@ func (l *testLogger) Write(level slog.Level, msg string, attrs ...any) {
 	l.logFn(l.formatMessage(msg, attrs...))
 }
 
-// TestOutputTestLogsRealtime verifies that test logs are output in real-time when outputTestLogs is enabled
-func TestOutputTestLogsRealtime(t *testing.T) {
+// TestOutputRealtimeLogs verifies that test logs are output in real-time when outputRealtimeLogs is enabled
+func TestOutputRealtimeLogs(t *testing.T) {
 	// Create a test file that outputs logs over time
 	testContent := []byte(`
 package feature_test
@@ -248,7 +248,7 @@ gates:
 	}
 
 	r := setupTestRunner(t, testContent, configContent)
-	r.outputTestLogs = true
+	r.outputRealtimeLogs = true
 	r.log = customLogger
 
 	// Run the test in a goroutine
@@ -293,8 +293,8 @@ gates:
 	}
 }
 
-// TestOutputTestLogsRealtimeDisabled verifies that test logs are output in real-time when outputTestLogs is disabled
-func TestOutputTestLogsRealtimeDisabled(t *testing.T) {
+// TestOutputRealtimeLogsDisabled verifies that test logs are output in real-time when outputRealtimeLogs is disabled
+func TestOutputRealtimeLogsDisabled(t *testing.T) {
 	// Create a test file that outputs logs over time
 	testContent := []byte(`
 package feature_test
@@ -336,7 +336,7 @@ gates:
 	}
 
 	r := setupTestRunner(t, testContent, configContent)
-	r.outputTestLogs = false
+	r.outputRealtimeLogs = false
 	r.log = customLogger
 
 	// Run the test in a goroutine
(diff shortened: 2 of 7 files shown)


scharissis:
Do you still think this is needed?

op-acceptor/Dockerfile
```diff
@@ -1,5 +1,8 @@
 FROM golang:1.23.5-alpine3.21 AS builder
 
+# Increase the proxy connect timeout to 60 seconds
+ENV GOPROXYCONNECTTIMEOUT=60s
```
scharissis: Does this actually exist? I can't find any mention of it in the docs.

op-acceptor/Dockerfile
```diff
@@ -9,7 +12,12 @@ COPY op-acceptor/ .
 
 RUN just build
 
-FROM alpine:3.21
+FROM golang:1.23.5-alpine3.21
+
+RUN apk add --no-cache ca-certificates build-base git
+
+# Increase the proxy connect timeout to 60 seconds
+ENV GOPROXYCONNECTTIMEOUT=60s
```
scharissis: Repeated above; rm?

op-acceptor/Dockerfile
```diff
@@ -30,7 +39,12 @@ chown -R app:app /go
 
 COPY --from=builder /app/bin/op-acceptor /app/
 COPY op-acceptor/example-validators.yaml /app/
+
+# Set ownership of the /app directory to allow the application to write logs and other files at runtime
+RUN chown -R app:app /app/
```
scharissis, line 38: Is this needed? The logs and files are working without this, I believe?

op-acceptor/runner/runner.go
```diff
@@ -415,6 +441,7 @@ func (r *runner) runAllTestsInPackage(ctx context.Context, metadata types.Valida
 }
 
 // listTestsInPackage returns all test names in a package
+// This may take a while to run as it needs to download all the dependencies
```
scharissis: This is no longer the case, I believe.

op-acceptor/runner/runner.go
```diff
@@ -379,7 +391,21 @@ func (r *runner) RunTest(ctx context.Context, metadata types.ValidatorMetadata)
...
+		fullPath := filepath.Join(r.workDir, metadata.Package)
+		if _, statErr := os.Stat(fullPath); os.IsNotExist(statErr) {
+			r.log.Error("Local package path does not exist, failing test", "validator", metadata.ID, "package", metadata.Package, "fullPath", fullPath)
+			return &types.TestResult{
+				Metadata: metadata,
+				Status:   types.TestStatusFail,
+				Error:    fmt.Errorf("local package path does not exist: %s", fullPath),
+			}, nil
+		}
+	}
```
scharissis, line 405: I haven't been able to find any evidence that this is needed.
I believe Go can and will resolve local paths, even when referenced as their full package paths.

Are you _sure_ it's needed? How?

op-acceptor/runner/runner.go
```diff
@@ -1161,16 +1208,14 @@ func (r *runner) testCommandContext(ctx context.Context, name string, arg ...str
...
 		}
 
-		env := append(
-			os.Environ(),
+		runEnv = append(runEnv,
 			// override the env URL with the one from the temp file
 			fmt.Sprintf("%s=%s", env.EnvURLVar, envFile.Name()),
 			// override the control resolution scheme with the original one
 			fmt.Sprintf("%s=%s", env.EnvCtrlVar, url.Scheme),
-			// Set DEVNET_EXPECT_PRECONDITIONS_MET=true to make tests fail instead of skip when preconditions are not met
```
scharissis, line 1170: why remove this?

op-acceptor/runner/runner.go
scharissis, line 405: I haven't been able to find any evidence that this is required. My understanding, and experience, suggest that it's not because Go finds local packages (even when referenced by their full package paths).

Are you sure it's needed?

op-acceptor/runner/runner_test.go
```diff
@@ -1583,3 +1583,46 @@ gates:
...
+func TestRunTest_PackagePath_Local(t *testing.T) {
+	r := setupDefaultTestRunner(t)
+
+	origPath := os.Getenv("PATH")
+	defer os.Setenv("PATH", origPath)
+
+	testCases := []struct {
+		name         string
+		packagePath  string
+		setupGit     bool // if false, PATH is emptied to simulate missing git
```
scharissis: I haven't seen an error that git is missing - where do you encounter this?
Does `-buildvcs=false` work around this perhaps?

op-acceptor/runner/runner.go
```diff
@@ -1144,7 +1187,11 @@ func (r *runner) testCommandContext(ctx context.Context, name string, arg ...str
 	cmd := exec.CommandContext(ctx, name, arg...)
 	cmd.Dir = r.workDir
 
+	// Always set the LOG_LEVEL environment variable
+	runEnv := append([]string{fmt.Sprintf("LOG_LEVEL=%s", r.testLogLevel)}, os.Environ()...)
```
scharissis: Is this supposed to be `TEST_LOG_LEVEL`?
Does it get honored?

scharissis (changes requested):

op-acceptor/Dockerfile
serpixel: This is needed if running the docker image locally unless we change the directory we run it on (like in CI).

op-acceptor/runner/runner.go
serpixel: Without this part if you have a test that uses a local path in 'package' and this path doesn't exist then the entire test suite will fail and it may not be obvious why.
We're using such paths in our test file so this would be relevant.

op-acceptor/runner/runner_test.go
serpixel: (Removed this), the test used to check for the presence of git on the Path. I added a check whilst trying to get the tests to run on dolphin but ended up adding the dependency on the image so this is no longer needed.

op-acceptor/runner/runner.go
serpixel: Restored it, thanks!

op-acceptor/runner/runner.go
serpixel: I added a comment above relating this, the tldr is that without this entry if you have a test that uses a local path in 'package' and this path doesn't exist then the entire test suite will fail.
We're using such paths in our test file so this would be relevant.

op-acceptor/Dockerfile
serpixel: Was supposed to be removed, back from when we were trying to get some logs on dolphin.

op-acceptor/runner/runner.go
serpixel: I have added a test for this to ensure the env variable is set. We need https://github.com/ethereum-optimism/optimism/pull/16050 too in order for the tests to be honoring the log level.

op-acceptor/runner/runner.go
scharissis: I'm not sure I follow, can you give me an example?

op-acceptor/runner/runner.go
scharissis: But is it used? I don't see where?

op-acceptor/runner/logging_test.go
```diff
@@ -68,3 +70,133 @@ gates:
...
+	var testResult2 *types.TestResult
+	for _, test := range result.Gates["logging-gate"].Suites["logging-suite"].Tests {
+		testResult2 = test
+		break
+	}
+	require.NotNil(t, testResult)
+
+	// Verify that logs are not captured when OutputTestLogs is disabled
+	assert.Equal(t, types.TestStatusPass, testResult2.Status)
+	assert.Contains(t, testResult2.Stdout, "This is a test output")
```
scharissis: Shouldn't this be asserting that it does NOT contain this output?

```suggestion
	// Verify that logs are not captured when OutputTestLogs is disabled
	assert.Equal(t, types.TestStatusPass, testResult2.Status)
	assert.NotContains(t, testResult2.Stdout, "This is a test output")
```

op-acceptor/runner/logging_test.go
```diff
@@ -68,3 +70,133 @@ gates:
 	assert.Contains(t, failingTest.Stdout, "This is some stdout output that should be captured")
 	assert.Contains(t, failingTest.Stdout, "This is a second line of output")
 }
+
+// TestLogLevelEnvironment verifies that the TEST_LOG_LEVEL environment variable is correctly set and used
+func TestLogLevelEnvironment(t *testing.T) {
```
scharissis, line 81: Let's also test the different levels behave differently?

scharissis (changes requested):

op-acceptor/runner/logging_test.go
serpixel: Actually the input is shown either way because we're not saving it to a file, that's what the test itself is testing. I think the other pre-existing tests cover the file logging part so I didn't add for that.

op-acceptor/runner/logging_test.go
serpixel: The test runner should be parsing the env variable (in the optimism repo) so I'm not sure we need to check the setup here

op-acceptor/Dockerfile
```diff
@@ -9,7 +9,9 @@ COPY op-acceptor/ .
 
 RUN just build
 
-FROM alpine:3.21
+FROM golang:1.23.5-alpine3.21
```
scharissis, line 12: I don't think we need the runner image to be updated, do we?

op-acceptor/runner/runner.go
```diff
@@ -563,10 +601,12 @@ func (r *runner) runSingleTest(ctx context.Context, metadata types.ValidatorMeta
 
 	// If we couldn't parse the output for some reason, create a minimal failing result
 	if parsedResult == nil {
+		r.log.Error("test exited with non-zero exit code", "exitCode", cmd.ProcessState.ExitCode())
```
scharissis, line 604: This line is confusing me; it need not necessarily have exited non-zero here?

op-acceptor/runner/logging_test.go
scharissis: I'm afraid I don't understand. How is this block of code verifying that logs are NOT captured?

```
	// Verify that logs are not captured when OutputTestLogs is disabled
	assert.Equal(t, types.TestStatusPass, testResult2.Status)
	assert.Contains(t, testResult2.Stdout, "This is a test output")
```

op-acceptor/Dockerfile
scharissis: spoke offline; "better safe than sorry" is the approach we're using for now.

scharissis (approved):
LGTM
As you suggested, it may be more self-evident if this flag is instead called `OutputRealtimeLogs` or `OutputLogStream`. But that's not a blocker for me.

PR #353, opened 2025-05-15
feat(op-acceptor): use static analysis to extract list of tests

sigma:
**Description**

The current approach of running `go test -list` has some drawbacks in
that it actually runs all the initialization code for the packages
that are dependencies of the tests.

This new approach simply parses the files that are in context for a
run and emits the list of test functions in there.

**Tests**

**Additional context**

**Metadata**

- Fixes #[Link to Issue]


op-acceptor/testlist/testlist.go
```diff
@@ -0,0 +1,87 @@
...
+	pkgDir := filepath.Join(workingDir, relPath)
+	entries, err := os.ReadDir(pkgDir)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read package directory: %w", err)
+	}
+
+	var testFunctions []string
+	fset := token.NewFileSet()
+
+	for _, entry := range entries {
```
janjakubnanista, line 61: We only do one level, no nested packages yeah?

janjakubnanista (approved):

op-acceptor/testlist/testlist.go
sigma: yep, that's what we do for now at least

PR #369, opened 2025-05-19
feat(proxyd): health-aware-load-balancing strategy for interop message validation

### LINKED ISSUE #363: proxyd: (interop-message-validation) health-aware load balancing
Currently, proxyd supports supervisor redundancy through the means of "multicall" interop_validation.strategy.

It operates by fanning out an incoming interop transaction to all the supervisor backends for its validation, and the validation succeeds if at least one supervisor responses with a success.

Though this provides HA guarantees for interop message validation, it does end up inducing additional undeserving load on the backing supervisors.

Another HA-oriented idea which can have lower impact on the backing supervisors is health-aware load-balancing between supervisors.

A/C:
- Estimate the effort and worthiness of having health-aware load balancing 
- Optionally, implement a "load-balanced" strategy under interop_validation which routes the incoming interop transactions to one healthy supervisor at a time in a round robin fashion.

yashvardhan-kukreja:

**Description**

Health-aware load-balancing of backing supervisors

**Tests**

New Tests added

**Additional context**

**Metadata**

- Fixes #363

### DIFF
--- a/proxyd/interop_strategy.go
+++ b/proxyd/interop_strategy.go
@@ -14,6 +14,8 @@ import (
 	"github.com/ethereum/go-ethereum/log"
 )
 
+const interopValidationStrategyKey = "interop_validation_strategy"
+
 type InteropStrategy interface {
 	Validate(ctx context.Context, req *RPCReq) error
 }
@@ -156,23 +158,8 @@ func (s *firstSupervisorStrategyImpl) Validate(ctx context.Context, req *RPCReq)
 
 	firstSupervisorUrl := s.urls[0]
 
-	httpCode, rpcErrorCode, err := performCheckAccessListOp(ctx, accessListToValidate, firstSupervisorUrl)
-
-	log.Debug(
-		"an interop validating backend has responded",
-		"supervisor_url", firstSupervisorUrl,
-		"req_id", GetReqID(ctx),
-		"method", "eth_sendRawTransaction",
-		"error", err,
-	)
-
-	rpcSupervisorChecksTotal.WithLabelValues(
-		firstSupervisorUrl,
-		httpCode,
-		rpcErrorCode,
-		string(FirstSupervisorStrategy),
-	).Inc()
-
+	ctx = context.WithValue(ctx, interopValidationStrategyKey, FirstSupervisorStrategy)
+	_, _, err = performCheckAccessListOp(ctx, accessListToValidate, firstSupervisorUrl)
 	return err
 }
 
@@ -203,23 +190,7 @@ func (s *multicallStrategyImpl) Validate(ctx context.Context, req *RPCReq) error
 		wg.Add(1)
 		go func(ctx context.Context, url string) {
 			defer wg.Done()
-			httpCode, rpcErrorCode, err := performCheckAccessListOp(ctx, accessListToValidate, url)
-
-			log.Debug(
-				"an interop validating backend has responded",
-				"supervisor_url", url,
-				"req_id", GetReqID(ctx),
-				"method", "eth_sendRawTransaction",
-				"error", err,
-			)
-
-			rpcSupervisorChecksTotal.WithLabelValues(
-				url,
-				httpCode,
-				rpcErrorCode,
-				string(MulticallStrategy),
-			).Inc()
-
+			_, _, err := performCheckAccessListOp(ctx, accessListToValidate, url)
 			resultChan <- err
 		}(ctx, url)
 	}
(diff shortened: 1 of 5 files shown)


proxyd/interop_strategy.go
```diff
@@ -251,6 +252,120 @@ func (s *multicallStrategyImpl) Validate(ctx context.Context, req *RPCReq) error
...
+		lastUnhealthy: time.Time{},
+	}
+}
+
+func (b *healthAwareBackend) IsHealthy() bool {
+	if b.lastUnhealthy.IsZero() {
+		return true
+	}
+
+	timeOneMinuteAgo := time.Now().Add(-10 * time.Second)
```
jelias2: This is 10s ago?

proxyd/interop_strategy.go
```diff
@@ -271,3 +386,60 @@ func performCheckAccessListOp(ctx context.Context, accessList []common.Hash, url
...
+	return len(b.backends)
+}
+
+func (b *loadBalancingBuffer) GetBackend() *healthAwareBackend {
+	b.mu.RLock()
+	defer b.mu.RUnlock()
+	return b.backends[b.currentBackendIndex]
+}
+
+// func (b *loadBalancingBuffer) RefreshIndexToHealthyBackend() {
```
jelias2: Delete if unused, or add description for why it needs to be in yet commented

yashvardhan-kukreja: Ohh yes, I forgot to delete it after commenting it out seeing its no use. Will do.

proxyd/interop_strategy.go
yashvardhan-kukreja: Ohh, I forgot to rename the variable name. 10s felt like more sane default than 1m. (don't wanna have a backoff implemented early on).

Will change the variable name

proxyd/interop_strategy.go
```diff
@@ -251,6 +252,120 @@ func (s *multicallStrategyImpl) Validate(ctx context.Context, req *RPCReq) error
...
+}
+
+func NewHealthAwareBackend(url string) *healthAwareBackend {
+	return &healthAwareBackend{
+		url:           url,
+		lastUnhealthy: time.Time{},
+	}
+}
+
+func (b *healthAwareBackend) IsHealthy() bool {
```
jelias2, line 297: It would be nice to mark backend healthiness with a prometheus metric. Mark the metric as 1 for unhealthy or 0 for healthy.

This way we can track which backends are health or not over time

yashvardhan-kukreja: Yep yep, will be tracking all the monitoring improvements here - https://github.com/ethereum-optimism/optimism/issues/15848?issue=ethereum-optimism%7Cinfra%7C365

proxyd/interop_strategy.go
```diff
@@ -251,6 +252,120 @@ func (s *multicallStrategyImpl) Validate(ctx context.Context, req *RPCReq) error
...
+	if b.lastUnhealthy.IsZero() {
+		return true
+	}
+
+	timeTenSecondsAgo := time.Now().Add(-10 * time.Second)
```
jelias2: This timeout should be configurable, and then update the tests as well to not use a 10s sleep

proxyd/interop_strategy.go
```diff
@@ -251,6 +252,120 @@ func (s *multicallStrategyImpl) Validate(ctx context.Context, req *RPCReq) error
...
+
+func (b *healthAwareBackend) MarkUnhealthy() {
+	b.lastUnhealthy = time.Now()
+}
+
+func (b *healthAwareBackend) Validate(ctx context.Context, accessList []common.Hash, req *RPCReq) (int, error) {
+	var outputHttpCode int
+	var outputErr error
+
+	httpCode, rpcErrorCode, err := performCheckAccessListOp(ctx, accessList, b.url)
```
jelias2: Any reason why performCheckAccessListOp converts and returns strings, but we convert them back into Ints later?

Should performCheckAccessListOp just return the ints?

proxyd/interop_strategy.go
yashvardhan-kukreja: Hmm, no strong reason to not do it apart from keeping it simple.

Can do

proxyd/interop_strategy.go
```diff
@@ -251,6 +252,120 @@ func (s *multicallStrategyImpl) Validate(ctx context.Context, req *RPCReq) error
...
+	)
+
+	rpcSupervisorChecksTotal.WithLabelValues(
+		b.url,
+		httpCode,
+		rpcErrorCode,
+		string(HealthAwareLoadBalancingStrategy),
+	).Inc()
+
+	outputHttpCode, parseErr := strconv.Atoi(httpCode)
```
jelias2: If performCheckAccessListOp returned ints this wouldn't be nessecary

proxyd/interop_strategy.go
yashvardhan-kukreja: No specific reason. Just that the output httpCode of performCheckAccessListOp was fed into the `WithLabelValues` function of prom metrics, which expects all strings.

No other strategy apart from load balancing one required the http code to be converted back to an int, so never got the problem surfaced.

jelias2 (commented):
Looking at the configuration for interop_strategies

```
[backends]
[backends.good]
rpc_url = "$GOOD_BACKEND_RPC_URL"
ws_url = "$GOOD_BACKEND_RPC_URL"

[backend_groups]
[backend_groups.main]
backends = ["good"]

[rpc_method_mappings]
eth_chainId = "main"
eth_sendRawTransaction = "main"

[interop_validation]
urls = [
"$VALIDATING_BACKEND_RPC_URL_1",
"$VALIDATING_BACKEND_RPC_URL_2"
]
strategy = "first-supervisor"
req_params_size_limit = 131072 ## 128KB
access_list_size_limit = 1000000

[interop_validation.sender_rate_limit]
allowed_chain_ids = [0, 420120003] # adding 0 allows pre-EIP-155 transactions
enabled = true
interval = "1s"
limit = 99999
```

the urls should be specificed if they are ws or rpc urls. Also it would be nice if the interop backends could be named and follow a similar declaration as backeds for the normal proxyd for consistency. That way metrics and logs don't use a full URLs

proxyd/interop_strategy.go
```diff
@@ -251,6 +252,120 @@ func (s *multicallStrategyImpl) Validate(ctx context.Context, req *RPCReq) error
...
+			continue
+		}
+
+		httpCode, err := backend.Validate(ctx, accessListToValidate, req)
+		if err == nil {
+			return nil
+		}
+
+		failedValidation := httpCode < 500
+		if failedValidation {
```
jelias2: nit, condense code

proxyd/interop_strategy.go
jelias2: Nit, I would recommend the main code logic uses ints all the way through, and then create a function specifically for the metrics that will convert to string if it really needs it

jelias2:
Overall looks solid, nice contribution and implementation, couple nits

jelias2 (approved):

PR #377, opened 2025-05-23
feat(op-acceptor): html result; add subtests.

scharissis:
Also show the subtests in the HTML output.

Screenshot 2025-05-23 at 19 52 50

### DIFF
--- a/op-acceptor/logging/filelogger.go
+++ b/op-acceptor/logging/filelogger.go
@@ -37,7 +37,6 @@ type FileLogger struct {
 	allLogsFile  string                // Path to the combined log file
 	mu           sync.Mutex            // Protects concurrent file operations
 	sinks        []ResultSink          // Collection of result consumers
-	resultBuffer []*types.TestResult   // Buffer for storing results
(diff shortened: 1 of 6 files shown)


op-acceptor/logging/filelogger.go
```diff
@@ -197,53 +201,46 @@ func (l *FileLogger) closeAllWriters() {
...
+	// Feed test result to all sinks
 	for _, sink := range l.sinks {
 		if err := sink.Consume(result, runID); err != nil {
-			return err
+			return fmt.Errorf("error in sink: %w", err)
 		}
 	}
 
+	// Buffer the result for potential summary operations
+	l.resultBuffer = append(l.resultBuffer, result)
```
janjakubnanista: Is the mutex lock/unlock no longer necessary here?

janjakubnanista (approved):

op-acceptor/logging/filelogger.go
scharissis: Glad you saw/commented; this was actually dead code I should have removed.
Thanks!

PR #380, opened 2025-05-28
feature(op-acceptor): improved json logging.

scharissis:
Raw JSON logs should now be created for all tests, regardless whether they passed or failed. Tidied up the code surrounding JSON sink access.

### DIFF
--- a/op-acceptor/README.md
+++ b/op-acceptor/README.md
@@ -133,16 +133,18 @@ go run cmd/main.go \
 
 For each test run, the following outputs are generated in the logs directory:
 
+0. HTML summary of the test run in `logs/results.html`
 1. Individual test log files in `logs/testrun-{run-id}/tests/`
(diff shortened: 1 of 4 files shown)


jelias2:
Typically for exporting logs we just print to std out. I wonder if its possible during each test, it would possible to just print the logs? 

This PR looks to be aggregating all logs into one sink, however is that sink actually printed?

scharissis:
> Typically for exporting logs we just print to std out. I wonder if its possible during each test, it would possible to just print the logs?
> 
> This PR looks to be aggregating all logs into one sink, however is that sink actually printed?

Separate to this PR we have recently introduced a flag to stream logs to stdout: https://github.com/ethereum-optimism/infra/pull/352

But what this PR does is capture all the JSON logs (`from go test -json`) and dumps them to one or more files. I don't think we want to stream these to stdout in particular.

jelias2 (approved):

PR #381, opened 2025-05-28
feat(op-acceptor): salt devstack keys generator

sigma:
**Description**

This requires https://github.com/ethereum-optimism/optimism/pull/16158 to have an actual effect.

At that point, the op-acceptor runID will be used to initialize the
keys generator, effectively scoping it to the run and getting it out of
the way of concurrent runs.

Also add a blurb in the logs to make sure we provide enough data to
reproduce these runs.

- needed for https://github.com/ethereum-optimism/optimism/issues/15885


scharissis (approved):
LGTM

PR #383, opened 2025-05-29
feat(op-acceptor): various reporting improvements

### LINKED ISSUE #187: [nat] Test timeouts
Handle test timeouts.

scharissis:
# Description

Various improvements to reporting, notably:
- New reporting package
- New ui package
- Intermediary test output format (TestTree) - shared by table and html outputs
- Improved table output
- Much improved html output
- Better timeout handling (tests that time out now report as such in the logs)
- Some code cleanup

# Images

Console output
Screenshot 2025-06-02 at 16 39 22

HTML output
Screenshot 2025-06-02 at 16 27 21

# Metadata

- Fixes https://github.com/ethereum-optimism/infra/issues/187


op-acceptor/runner/runner.go
```diff
@@ -463,6 +463,10 @@ func (r *runner) runTestList(ctx context.Context, metadata types.ValidatorMetada
 	testResults := make(map[string]*types.TestResult)
 	var failedTestsStdout strings.Builder
 	var aggregatedRawJSON []byte // Store aggregated raw JSON for the test list
+	var timeoutCount int         // Track how many tests timed out
+	var timedOutTests []string   // Track which tests timed out
```
serpixel: I would personally initialise it, append should work anyways but other operations may end up throwing an error.
```suggestion
	var timedOutTests = make([]string, 0)   // Track which tests timed out
```

op-acceptor/runner/runner.go
```diff
@@ -507,23 +528,62 @@ func (r *runner) runTestList(ctx context.Context, metadata types.ValidatorMetada
...
-		Error:    errors.Join(testErrors...),
+		Error:    finalError,
 		Duration: totalDuration,
 		SubTests: testResults,
 		Stdout:   failedStdout,
-	}, nil
+	}
+
+	// Log the package result
+	if r.fileLogger != nil {
```
serpixel, line 579: Shouldn't we log this anyways even if the fileLogger is null?

op-acceptor/runner/runner.go
```diff
@@ -580,20 +645,82 @@ func (r *runner) runSingleTest(ctx context.Context, metadata types.ValidatorMeta
...
+		if stderr.Len() > 0 {
+			if result.Error != nil {
+				result.Error = fmt.Errorf("%w\nstderr: %s", result.Error, stderr.String())
+			} else {
+				result.Error = fmt.Errorf("timeout stderr: %s", stderr.String())
+			}
+		}
+
+		// Force logging of timeout result to ensure it's captured
+		if r.fileLogger != nil {
```
serpixel, line 719: Same here, we may want this on the console

op-acceptor/runner/runner_test.go
```diff
@@ -1619,3 +1622,132 @@ func TestRunTest_PackagePath_Local(t *testing.T) {
...
+}
+
+func TestAnotherSlowTimeout(t *testing.T) {
+	// This test will also timeout
+	time.Sleep(5 * time.Second)
+}
+
+func TestAnotherFast(t *testing.T) {
+	// This test passes quickly
+}
```
serpixel, line 1658: May be worth adding a test is marked as skip here. Mostly because we will skip tests when the network is not having a compatible setup (e.g. if we require an interop network but have a simple network).

op-acceptor/reporting/formatters.go
```diff
@@ -0,0 +1,562 @@
...
+			if tf.showIndividualTests {
+				tf.addTestsWithSubtests(t, suite.Tests, true)
+			}
+		}
+
+		t.AppendSeparator()
+	}
+
+	// Update the table style based on overall result status
+	if data.HasFailures {
```
serpixel: I am wondering whether we need to add some logic here to handle non-TTY / cases where the env NO_COLOR is set to true.

op-acceptor/runner/runner.go
```diff
@@ -471,6 +475,8 @@ func (r *runner) runTestList(ctx context.Context, metadata types.ValidatorMetada
 		testMetadata.RunAll = false
 		testMetadata.FuncName = testName
 
+		r.log.Debug("Running individual test in package", "test", testName, "package", metadata.Package)
+
 		// Run the individual test
 		testResult, err := r.runSingleTest(ctx, testMetadata)
```
serpixel, line 481: Outside of the scope of this PR. Should we be handling the case where the test OOM's / set some limits for the test process?

op-acceptor/reporting/formatters.go
```diff
@@ -0,0 +1,562 @@
...
+
+	return &HTMLFormatter{
+		template: tmpl,
+	}, nil
+}
+
+// Format formats the report data as HTML
+func (hf *HTMLFormatter) Format(data *ReportData) (string, error) {
+	// Convert ReportData to the structure expected by the HTML template
+	htmlData := &HTMLSummaryData{
```
serpixel: One thing that occurred to me when checking another PR. We could potentially dump the chain setup / status with PrintChain() and then embed that into the results (outside of scope for this PR).

op-acceptor/runner/runner.go
scharissis: Done

op-acceptor/runner/runner.go
scharissis: Done; added a `r.Log.Info`

op-acceptor/runner/runner.go
scharissis: Done; added a `r.Log.Info`

op-acceptor/runner/runner_test.go
scharissis: Good call; done.

op-acceptor/reporting/formatters.go
scharissis: The library we're using (github.com/jedib0t/go-pretty/v6/table) should handle this for us.

op-acceptor/reporting/formatters.go
scharissis: Created https://github.com/ethereum-optimism/infra/issues/386
Feel free to edit/elaborate in the ticket.
Thanks!

op-acceptor/reporting/html_sink.go
```diff
@@ -0,0 +1,195 @@
...
+	// Create file writer and report generator
+	writer := NewFileWriter(htmlFile)
+	generator := NewReportGenerator(s.builder, s.formatter, writer)
+
+	// Generate the report
+	return generator.GenerateReport(reportData)
+}
+
+// ReportingTextSummarySink is a modern text summary sink that uses the unified reporting structure
+type ReportingTextSummarySink struct {
```
sigma: I think it'd be less confusing to put the non-html parts in a separate file

op-acceptor/runner/runner.go
```diff
@@ -480,6 +486,16 @@ func (r *runner) runTestList(ctx context.Context, metadata types.ValidatorMetada
 		testResults[testName] = testResult
 		totalDuration += testResult.Duration
 
+		// Check if this test timed out
+		if testResult.TimedOut {
+			timeoutCount++
```
sigma, line 492: nit: not really important right now, but we should probably make sure the counter increments are atomic to prevent data races if and when we have parallel runs

op-acceptor/logging/templates/results.tmpl.html
```diff
@@ -9,227 +9,728 @@
...
+                {{if .Gate}}<span>Gate: {{.Gate}}</span>{{end}}
+                {{if .Suite}}<span>Suite: {{.Suite}}</span>{{end}}
+                <span class="duration">{{.DurationFormatted}}</span>
+                {{if .LogPath}}<a href="{{.LogPath}}" target="_blank" class="log-link">View Log</a>{{end}}
+            </div>
+        </div>
+        {{end}}
+    </div>
 
     <script>
```
sigma: that's a lot of inline script now. Should we split this into a .js file ?

op-acceptor/reporting/formatters.go
```diff
@@ -0,0 +1,562 @@
...
+
+// generateTreePrefix creates tree-style prefixes for hierarchical display
+func generateTreePrefix(level int, isSubTest bool) string {
+	switch level {
+	case 0: // Gate level
+		return ""
+	case 1: // Suite level
+		return "├── "
+	case 2: // Test level
+		if isSubTest {
```
sigma: this feels like a small issue.
In theory you can have any depth of subtest nesting, so we should probably just have a numeric depth information that we use to compute the tree symbols

op-acceptor/reporting/formatters.go
```diff
@@ -0,0 +1,562 @@
...
+			parentTests = append(parentTests, test)
+		}
+	}
+
+	// Display parent tests followed by their subtests
+	for i, test := range parentTests {
+		// Determine prefix for main test
+		var prefix string
+		if isInSuite {
+			prefix = "│   ├──"
```
sigma: I'd propose we move those ASCII fragments into constants, and compose them instead of having static strings throughout the code

op-acceptor/reporting/formatters.go
```diff
@@ -0,0 +1,562 @@
...
+		fmt.Fprintf(&summary, "=================\n")
+
+		for _, gate := range data.Gates {
+			fmt.Fprintf(&summary, "Gate: %s (%s) [%s]\n", gate.Name, formatDuration(gate.Duration), strings.ToUpper(string(gate.Status)))
+
+			for _, suite := range gate.Suites {
+				fmt.Fprintf(&summary, "  Suite: %s (%s) [%s]\n", suite.Name, formatDuration(suite.Duration), strings.ToUpper(string(suite.Status)))
+
+				for _, test := range suite.Tests {
+					prefix := "    "
```
sigma: similarly for the spaces, we should probably compute them based on the situation.

I think that's where an intermediate representation of the tree could be used to factor the handling across html/txt backends

op-acceptor/reporting/report.go
```diff
@@ -0,0 +1,647 @@
...
+	}
+	return time.Time{}, false
+}
+
+// createTestItem creates a ReportTestItem from a TestResult
+func (rb *ReportBuilder) createTestItem(testResult *types.TestResult, isSubTest bool, parentTest string, level int, executionOrder int) ReportTestItem {
+	name := testResult.Metadata.FuncName
+
+	// Handle named subtests (e.g., "TestChainFork/Chain_0" -> "Chain_0")
+	if isSubTest && strings.Contains(name, "/") {
```
sigma: 2 things here:
- I think that's the other place where arbitrary-level subtests should be handled too
- we should also probably abstract away the Go-specific logic (slash-separated names) in anticipation of more test kinds popping up at some point.

scharissis: Arbitrary subtests should be handled now.

But the non-Go tests; I don't thinl we should address that in this PR - would you disagree?

op-acceptor/reporting/formatters.go
scharissis: Spaces handled.
Will investigate an intermediate representation, but probably outside of this PR. It keeps growing ... :)

op-acceptor/reporting/formatters.go
scharissis: Done; added a 'ui' package with this.

op-acceptor/reporting/formatters.go
scharissis: Refactored some of this nesting code.

op-acceptor/logging/templates/results.tmpl.html
scharissis: Agreed. Done.

op-acceptor/reporting/html_sink.go
scharissis: Agreed, done, thanks.

op-acceptor/reporting/report.go
sigma: that's fine. a TODO would be plenty :)

op-acceptor/reporting/tree_formatters.go
```diff
@@ -0,0 +1,607 @@
...
+			case 0:
+				return "indent-0"
+			case 1:
+				return "indent-1"
+			case 2:
+				return "indent-2"
+			case 3:
+				return "indent-3"
+			default:
+				return fmt.Sprintf("indent-%d", depth)
```
sigma: I guess we can just use this directly for all depths :)

op-acceptor/reporting/tree_formatters.go
```diff
@@ -0,0 +1,607 @@
...
+			}
+		},
+		"multiply": func(a, b int) int {
```
sigma, line 114: what's this for?

op-acceptor/reporting/tree_formatters.go
```diff
@@ -0,0 +1,607 @@
...
+// NewTreeHTMLFormatter creates a new tree-based HTML formatter
+func NewTreeHTMLFormatter(templateContent string) (*TreeHTMLFormatter, error) {
+	tmpl, err := template.New("tree-report").Funcs(template.FuncMap{
+		"formatDuration": func(d time.Duration) string {
+			if d < time.Second {
+				return fmt.Sprintf("%dms", d.Milliseconds())
+			}
+			return d.Truncate(time.Millisecond).String()
+		},
+		"getStatusClass": func(status types.TestStatus) string {
```
sigma, line 105: those 2 functions (getStatusClass, getStatusText) probably should be consolidated with getStatusString below

op-acceptor/reporting/tree_formatters.go
```diff
@@ -0,0 +1,607 @@
...
+	nodeData := map[string]interface{}{
+		"id":       node.ID,
+		"name":     node.Name,
+		"type":     node.Type,
+		"status":   node.Status,
+		"duration": node.Duration,
+		"depth":    node.Depth,
+	}
+
+	if node.Package != "" {
```
sigma: I'm not sure I completely understand all this. Presumably there's not a whole lot of difference between an  absent key and an empty string, so why are we making one here? (and above)

Wouldn't it be clearer to have a proper object to marshal (with the right "omitempty" and whatnot), vs having those maps with optional items?

op-acceptor/ui/tree.go
```diff
@@ -0,0 +1,126 @@
...
+	// Composite patterns for common use cases
+	TreeSubTestBranch     = "│       ├──"         // For subtests under tests
+	TreeSubTestLastBranch = "│       └──"         // For last subtest under tests
+	TreeSubTestError      = "│       │       └──" // For errors under subtests
+
+	// Suite-specific patterns (with additional indentation)
+	SuiteBranch        = "    ├── "                // Branch under suite
+	SuiteLastBranch    = "    └── "                // Last branch under suite
+	SuiteContinue      = "    │   "                // Continue line under suite
+	SuiteSubTestBranch = "    │       ├──"         // Subtest under suite test
```
sigma: I haven't looked at the details, but the existence of this constant seems to suggest we're not going to be able to represent deeper levels of subtests still?

op-acceptor/ui/tree_test.go
```diff
@@ -0,0 +1,330 @@
+package ui
+
+import (
+	"strings"
+	"testing"
+	"unicode/utf8"
+)
+
+func TestTreeConstants(t *testing.T) {
```
sigma: that seems like an odd thing to test :)

op-acceptor/README.md
```diff
@@ -131,19 +131,74 @@ go run cmd/main.go \
...
+   - Structured format for easy reading
+   - Contains metadata for each test (package, gate, suite, etc.)
+
+5. **Raw Events Log** (`raw_go_events.log`)
+   - Complete raw JSON output from all tests
+   - Compatible with `go test -json` format
+   - Useful for integration with Go test analysis tools
+   - Can be processed by tools like `gotestsum`
+
+#### Test Log Naming
```
serpixel, line 184: Not for now, but a possible improvement would be to add a runid or timestamp like other test frameworks do.

op-acceptor/ui/tree_test.go
scharissis: Removed! :)

op-acceptor/reporting/tree_formatters.go
scharissis: Done

op-acceptor/reporting/tree_formatters.go
scharissis: No longer used. Removed this and one other. Thanks!

op-acceptor/reporting/tree_formatters.go
scharissis: Good call; done. I've significantly simplified things here.

op-acceptor/reporting/tree_formatters.go
scharissis: I've changed it to tagged objects with JSON marshalling; seems a bit cleaner indeed. Thanks!

op-acceptor/ui/tree.go
scharissis: When I re-checked this, I found that, indeed, the data layer supported it but not the presentation layer. Updated it and added a test as well.

op-acceptor/ui/tree.go
```diff
@@ -0,0 +1,118 @@
...
+	BoxTopLeft     = "┌"
+	BoxTopRight    = "┐"
+	BoxBottomLeft  = "└"
+	BoxBottomRight = "┘"
+	BoxVertical    = "│"
+	BoxHorizontal  = "─"
+	BoxTeeRight    = "├"
+	BoxTeeLeft     = "┤"
+
+	// Suite-specific patterns (with additional indentation)
```
sigma, line 31: nit: in theory these shouldn't be needed, as they're composite symbols (indent + tree connector)
Removing them would avoid having use-case constructs (suites) creep into this package.
Anyways, no big deal :)

sigma (approved):

serpixel (approved):

op-acceptor/ui/tree.go
scharissis: Yeah. Will address in a follow-up!
Thanks! <3

PR #387, opened 2025-05-31
Fix broken link to op-acceptor README in main README

GarmashAlex:
Corrected the path to the op-acceptor component's README file in the main README. The previous link pointed to a non-existent acceptor/README.md; it now correctly points to op-acceptor/README.md. This resolves the broken link error and ensures users can access the relevant documentation.


scharissis (approved):
Thanks!

PR #405, opened 2025-06-10
feat(proxyd): don't use semaphore if allowing unlimited concurrency

charlesma4:
**Description**

Instead of creating semaphore with unlimited weight, just skip using semaphore entirely if we allow unlimited concurrent requests. Also adds a new log on init that explicitly tells us if we're using unlimited or limited concurrent requests.

**Tests**

Unit test to make sure unlimited vs. limited requests works as expected

**Additional context**

Unichain proxyd previously had issues where requests were incorrectly being rate limited due to semaphore exhaustion, despite unlimited concurrency being configured. The RC is still unclear, but having explicit logs on init, and only using the semaphore when necessary will help with future debugging.

### DIFF
--- a/proxyd/proxyd.go
+++ b/proxyd/proxyd.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
+	"math"
 	"net/http"
 	"os"
 	"time"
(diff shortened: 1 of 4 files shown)


jelias2:
The change looks fine. However I would suggest adding a new server conifg field/option such as `config.server.disableConcurrentSemaphoreRPC` which defaults to false to toggle the behavior, instead of 0 for maxConcurrentRPCRequests

This way the new behavior is hidden behind a flag, and the default behavior isn't updated immediately

charlesma4:
Unable to repro integ test failure, is the test flaky?

image

proxyd/proxyd.go
```diff
@@ -112,10 +111,15 @@ func Start(config *Config) (*Server, func(), error) {
 	}
 
 	maxConcurrentRPCs := config.Server.MaxConcurrentRPCs
-	if maxConcurrentRPCs == 0 {
-		maxConcurrentRPCs = math.MaxInt64
+	var rpcRequestSemaphore *semaphore.Weighted
```
jelias2, line 115:
```

if config.Server.DisableConcurrentRequestSemaphore {
	rpcRequestSemaphore = nil
	log.Info("Using disabling concurrent request semaphore ")
} else {
	if maxConcurrentRPCs == 0 {
		maxConcurrentRPCs = math.MaxInt64
	}
	rpcRequestSemaphore = semaphore.NewWeighted(maxConcurrentRPCs)
	log.Info("Using max concurrent RPCs", "maxConcurrentRPCs", maxConcurrentRPCs)
}
```

jelias2:
> Unable to repro integ test failure, is the test flaky?
> 
> image

Yes that multicall test has been known to flake

jelias2 (approved):

PR #411, opened 2025-06-23
feat(op-acceptor): first-class orchestrator support.

scharissis:
Added support for both the op-devstack orchestrator modes, allowing op-acceptor to run in-memory tests as well.

### DIFF
--- a/op-acceptor/cmd/main.go
+++ b/op-acceptor/cmd/main.go
@@ -19,7 +19,7 @@ import (
 )
 
 var (
-	Version   = "v0.1.0"
+	Version   = "v0.5.0"
 	GitCommit = ""
 	GitDate   = ""
 )
--- a/op-acceptor/nat.go
+++ b/op-acceptor/nat.go
@@ -20,6 +20,7 @@ import (
 
 	"github.com/ethereum-optimism/infra/op-acceptor/addons"
 	"github.com/ethereum-optimism/infra/op-acceptor/exitcodes"
+	"github.com/ethereum-optimism/infra/op-acceptor/flags"
 	"github.com/ethereum-optimism/infra/op-acceptor/logging"
 	"github.com/ethereum-optimism/infra/op-acceptor/metrics"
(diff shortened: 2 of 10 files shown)


op-acceptor/cmd/main.go
```diff
@@ -19,7 +19,7 @@ import (
 )
 
 var (
-	Version   = "v0.1.0"
+	Version   = "v0.5.0"
```
scharissis, line 22: we don't typically bother and I don't think it matters. But I noticed it, so I bumped it. 0.5.0 will be the next release as these are potentially breaking changes

op-acceptor/nat.go
```diff
@@ -69,17 +70,38 @@ func New(ctx context.Context, config *Config, version string, shutdownCallback f
...
+	var devnetEnv *env.DevnetEnv
+	var networkName string
 
-	devnetEnv, err := env.LoadDevnetFromURL(envURL)
-	if err != nil {
-		return nil, fmt.Errorf("failed to load devnet environment from %s: %w", envURL, err)
+	// Handle different orchestrator types
+	switch config.Orchestrator {
+	case flags.OrchestratorSysext:
+		// For sysext, we need DEVNET_ENV_URL and load the devnet environment
```
teddyknox: Could we incorporate this into the CLI as well?

teddyknox (commented):
This is a great direction to be going in, but we could take it further by incorporating a CLI param for `--devnet-env-url` so that the user does not need to set the DEVNET_ENV_URL env var manually.

op-acceptor/nat.go
scharissis: Sure thing. I had intended to add this in a subsequent PR, but I'll add it here/now.

scharissis: This has now been done; PTAL.

serpixel (approved):

teddyknox (approved):

op-acceptor/nat.go
teddyknox: Thanks!

PR #420, opened 2025-07-07
feat(proxyd): 499 error code for context canceled

charlesma4:
**Description**
Introducing artificial "499" error code to be included in metrics, for when client cancels the request. This is useful for monitoring, and helps distinguish between genuine server issues (500) vs. client-side issues.

[Cloudflare](https://developers.cloudflare.com/support/troubleshooting/http-status-codes/4xx-client-error/error-499/) already uses this error code to indicate client cancellations.

- new `ErrContextCanceled` error with 499 error code
- bubble up context canceled to correctly be included in httpErrorCodesTotal metric

**Tests**

Unit test to make sure new the metric count is incrementing correctly when 499 is seen.

### DIFF
--- a/proxyd/backend.go
+++ b/proxyd/backend.go
@@ -594,6 +594,11 @@ func (b *Backend) Forward(ctx context.Context, reqs []*RPCReq, isBatch bool) ([]
 			)
 		case ErrContextCanceled:
 			// return immediately on client cancellation
+			log.Debug("context canceled while forwarding request",
+				"name", b.Name,
+				"req_id", GetReqID(ctx),
+				"err", err,
+			)
 			return nil, err
 		default:
 			lastError = err
@@ -1518,7 +1523,7 @@ func (c *LimitedHTTPClient) DoLimited(req *http.Request) (*http.Response, error)
 			return nil, ErrContextCanceled
 		}
 		tooManyRequestErrorsTotal.WithLabelValues(c.backendName).Inc()
-		return nil, ErrTooManyRequests
+		return nil, wrapErr(err, ErrTooManyRequests.Message)
 	}
 	defer c.sem.Release(1)
 	return c.Do(req)
--- a/proxyd/backend.go
+++ b/proxyd/backend.go
@@ -119,6 +119,12 @@ var (
 		HTTPErrorCode: 413,
 	}
 
+	ErrContextCanceled = &RPCErr{
+		Code:          JSONRPCErrorInternal - 23,
+		Message:       "client disconnected",
+		HTTPErrorCode: 499,
+	}
+
 	ErrBackendUnexpectedJSONRPC = errors.New("backend returned an unexpected JSON-RPC response")
 
 	ErrConsensusGetReceiptsCantBeBatched = errors.New("consensus_getReceipts cannot be batched")
@@ -580,6 +586,9 @@ func (b *Backend) Forward(ctx context.Context, reqs []*RPCReq, isBatch bool) ([]
 				"req_id", GetReqID(ctx),
 				"err", err,
 			)
+		case ErrContextCanceled:
+			// return immediately on client cancellation
+			return nil, err
 		default:
 			lastError = err
 			log.Warn(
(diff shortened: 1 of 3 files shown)


proxyd/errors.go
```diff
@@ -1,6 +1,11 @@
 package proxyd
 
-import "fmt"
+import (
+	"errors"
+	"fmt"
+)
+
+var ErrTooManyRequests = errors.New("too many requests")
```
jelias2: Any reason this is placed in errors.go? Most of the errors are defined in backend.go, would recommend defining this there, unless there is a specific reason it should be in errors.go

jelias2: Also it looks like ErrToManyRequests could also be used in the [DoLimitedFunction](https://github.com/ethereum-optimism/infra/blob/7d490c17d1867b3c875f4c5eadf0db0c49bd5cef/proxyd/backend.go?plain=1#L1498)

proxyd/backend.go
```diff
@@ -746,8 +755,13 @@ func (b *Backend) doForward(ctx context.Context, rpcReqs []*RPCReq, isBatch bool
 	start := time.Now()
 	httpRes, err := b.client.DoLimited(httpReq)
 	if err != nil {
-		b.intermittentErrorsSlidingWindow.Incr()
-		RecordBackendNetworkErrorRateSlidingWindow(b, b.ErrorRate())
+		if !(errors.Is(err, context.Canceled) || errors.Is(err, ErrTooManyRequests)) {
```
jelias2, line 769: Should the second half of the conditional also contain a negation?
```suggestion
		if !(errors.Is(err, context.Canceled) || !errors.Is(err, ErrTooManyRequests)) {
```

charlesma4: I think the current logic is correct

if it's untrue that it's either context canceled or too many requests, then it's a real error and we should track it

the negation is for both errors already

proxyd/backend.go
```diff
@@ -1605,6 +1619,20 @@ func (bg *BackendGroup) ForwardRequestToBackendGroup(
...
 				}
 			}
+			// context canceled happens when either the client cancels the request
+			// or proxyd cancels the request. Proxyd only cancels requests when
+			// the server is shutting down, so this must be the client cancelling
+			// the request.
+			//
+			// We catch this error here so that we don't mark this request as
+			// unserviceable
+			if errors.Is(err, ErrContextCanceled) {
```
jelias2: On lines 1596 - 1608. There is already a conditional which returns and empty RPCResponse and the error type. Could this conditional be added there? 

Why doesn't the existing context.Canceled error check catch this?

```
			if errors.Is(err, ErrConsensusGetReceiptsCantBeBatched) ||
				errors.Is(err, ErrConsensusGetReceiptsInvalidTarget) ||
				// context canceled happens when either the client cancels the request
				// or proxyd cancels the request. Proxyd only cancels requests when
				// the server is shutting down, so this must be the client cancelling
				// the request.
				errors.Is(err, context.Canceled) ||
				errors.Is(err, ErrContextCanceled) 
				 {
				return &BackendGroupRPCResponse{
					RPCRes:   nil,
					ServedBy: "",
					error:    err,
				}
			}
```

proxyd/errors.go
charlesma4: copied from our local fork, will move it over 🫡

proxyd/backend.go
charlesma4: ah using the existing one is cleaner, will update

proxyd/backend.go
```diff
@@ -1494,8 +1514,11 @@ func (c *LimitedHTTPClient) DoLimited(req *http.Request) (*http.Response, error)
 	}
 
 	if err := c.sem.Acquire(req.Context(), 1); err != nil {
+		if errors.Is(err, context.Canceled) {
+			return nil, ErrContextCanceled
+		}
 		tooManyRequestErrorsTotal.WithLabelValues(c.backendName).Inc()
-		return nil, wrapErr(err, "too many requests")
+		return nil, ErrTooManyRequests
```
jelias2: Should the code should still wrap the error on line 1521?

```
		return nil, wrapErr(err, ErrToManyRequests.Message())
```

proxyd/backend.go
```diff
@@ -580,6 +592,9 @@ func (b *Backend) Forward(ctx context.Context, reqs []*RPCReq, isBatch bool) ([]
 				"req_id", GetReqID(ctx),
 				"err", err,
 			)
+		case ErrContextCanceled:
+			// return immediately on client cancellation
```
jelias2, line 596: Can a debug log be added here to emit that the context was cancelled

jelias2 (approved):

PR #424, opened 2025-07-15
Enable public access and authentication

cody-wang-cb:
**Description**
This is useful when switching from API gated to public enabled access, so that it won't break existing users who have API keys in their software.

**Tests**
Added an integration test for this

**Metadata**

- Fixes #[Link to Issue]


cody-wang-cb:
@jelias2 @scharissis hey guys can I get a quick review on this, just upstreaming something we needed in the past

proxyd/integration_tests/public_access_test.go
```diff
@@ -0,0 +1,40 @@
...
+import (
+	"net/http"
+	"os"
+	"testing"
+
+	"github.com/ethereum-optimism/infra/proxyd"
+	"github.com/stretchr/testify/require"
+)
+
+func TestPublicAccess(t *testing.T) {
```
scharissis, line 13: Lets also cover the case where we reject invalid authentication requests, even though public mode is enabled.

scharissis:
Tests and Lint are failing in CI.
Are you able to replicate and fix locally?

proxyd/integration_tests/public_access_test.go
cody-wang-cb: Just added

cody-wang-cb:
@scharissis wanna take a look at this PR again when you have a chance

jelias2 (approved):

proxyd/integration_tests/public_access_test.go
scharissis: Thanks

PR #425, opened 2025-07-16
fix(op-acceptor): zero durations on failed runs.

scharissis:
Test runs should now show a non-zero total duration regardless of result status.

BEFORE:
Screenshot 2025-07-16 at 14 12 58

AFTER:
Screenshot 2025-07-16 at 14 13 08

### DIFF
--- a/op-acceptor/runner/runner.go
+++ b/op-acceptor/runner/runner.go
@@ -856,8 +856,7 @@ func isMainTestEvent(event TestEvent, mainTestName string) bool {
 }
 
 // processMainTestEvent handles events for the main test
-func processMainTestEvent(event TestEvent, result *types.TestResult, testStart, testEnd *time.Time,
-	errorMsg *strings.Builder, hasSkip *bool) {
+func processMainTestEvent(event TestEvent, result *types.TestResult, testStart, testEnd *time.Time, errorMsg *strings.Builder, hasSkip *bool) {
 	switch event.Action {
 	case ActionStart:
 		*testStart = event.Time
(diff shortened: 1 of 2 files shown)


op-acceptor/runner/runner.go
```diff
@@ -867,14 +866,19 @@ func processMainTestEvent(event TestEvent, result *types.TestResult, testStart,
 	case ActionFail:
 		*testEnd = event.Time
 		result.Status = types.TestStatusFail
+		// If we have an elapsed time from the event, use it as a fallback
+		if event.Elapsed > 0 && result.Duration == 0 {
+			result.Duration = time.Duration(event.Elapsed * float64(time.Second))
```
janjakubnanista, line 871: Why plus one second here?

op-acceptor/runner/runner.go
```diff
@@ -931,6 +935,11 @@ func calculateSubTestDuration(subTest *types.TestResult, event TestEvent, subTes
 		// Fallback to elapsed if provided
 		subTest.Duration = time.Duration(event.Elapsed * float64(time.Second))
 	}
+
+	// If we still don't have a duration and this is a failed test, try to use the elapsed time
+	if subTest.Duration == 0 && subTest.Status == types.TestStatusFail && event.Elapsed > 0 {
+		subTest.Duration = time.Duration(event.Elapsed * float64(time.Second))
```
janjakubnanista, line 941: Same here

op-acceptor/runner/runner.go
scharissis: I'm not adding a second, I'm simply converting `event.Elapsed` to a time.Duration and then using that as our `result.Duration`.

```
type TestEvent struct {
	...
	Elapsed float64   // Elapsed time in seconds for the specific action
}
```

janjakubnanista: Ahhhhh 🤦 it's `*` not `+` sorry

janjakubnanista (approved):

op-acceptor/runner/runner.go
scharissis: No problem, thanks for the review!

PR #428, opened 2025-07-21
feat(op-acceptor): gateless mode

scharissis:
Introduce a gateless mode which tries to find and run all tests within a directory.

Running all tests, in sysgo mode:
Screenshot 2025-07-23 at 16 31 36

### DIFF
--- a/op-acceptor/README.md
+++ b/op-acceptor/README.md
@@ -1,19 +1,95 @@
 # Network Acceptance Tester (op-acceptor)
+
 op-acceptor is a tool to run checks against a network to determine if it is ready for production. It helps ensure your network meets all necessary requirements before deployment.
 The checks it runs are standard Go tests.
+
+## Usage Modes
+
+### Gateless Mode
+
+For quick testing without configuration, op-acceptor can auto-discover and run all tests in a directory:
+
+```bash
+# Run all tests in a directory
+op-acceptor --testdir ./my-acceptance-tests
+
+# Run all tests recursively (supports Go's ... notation)
+op-acceptor --testdir ./my-acceptance-tests/...
+
+# With additional options
+op-acceptor --testdir ./my-tests \
+    --orchestrator sysgo \
+    --run-interval 0 \
+    --log.level info
+```
+
+In gateless mode:
+- **No validators config required**: Tests are auto-discovered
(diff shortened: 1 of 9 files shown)


op-acceptor/README.md
```diff
@@ -1,20 +1,79 @@
...
+### Gateless Mode
+
+For quick testing without configuration, op-acceptor can auto-discover and run all tests in a directory:
+
+```bash
+# Run all tests in a directory
+op-acceptor --testdir ./my-acceptance-tests
+
+# Run all tests recursively (supports Go's ... notation)
+op-acceptor --testdir ./my-acceptance-tests/...
```
serpixel, line 17: Would be great if we supported a syntax similar to go test, e.g. `op-acceptor ./my-acceptance-tests/...`

op-acceptor/config.go
```diff
@@ -17,6 +17,7 @@ type Config struct {
 	TestDir            string
 	ValidatorConfig    string
 	TargetGate         string
+	GatelessMode       bool // New field to indicate gateless mode
```
serpixel, line 20: We would need to add an option to configure the timeout in this case, as we would normally configure it through the validator yaml.

serpixel (commented):
Left a couple of comments if you could check 👍

op-acceptor/README.md
scharissis: Do you mean without needing to provide `--testdir`?
Because this is similar to go test otherwise? `op-acceptor --testdir ./my-acceptance-tests/...`

op-acceptor/config.go
scharissis: Good point. I'll cook on this.

scharissis: Added a new '--timeout', so now we have:
* --default-timeout = per test timeout
* --timeout = overall run timeout

op-acceptor/README.md
serpixel: Yes exactly, without `--testdir`

serpixel (approved):

op-acceptor/README.md
scharissis: I'll look into it (different PR).
Thanks!

PR #436, opened 2025-07-25
Proxyd: Allow backend ResponseTimeout to be set in ms + to be configurable per backend

dpulitano:
**Description**

Rollups often need certain RPC calls to be fast (i.e. faster than 1s). This PR adds a new configuration field intended to replace `ResponseTimeoutSeconds`: `ResponseTimeoutMilliseconds` that allows response time to be tuned more finely

In order to allow this to be configured per RPC method, I had to allow the overriding of the global responsetimeout on a per-backend basis. It is useful to also allow overriding maxRetries, so added that as well.

Finally, I am adding a fix where a backoff was added in even if it's the last or only retry for a backend. I am changing this to only add a backoff if there are more retries pending for that backend.

**Tests**

- smoke testing ✅ 
- having eth_getBlockByNumber point to a new backend group with:
    - the primary backend having a timeout of 2ms, and 
    - a set of fallbacks
    => 100% client errors, and a fallback being attempted

Integration and unit tests added

**Additional context**

**Metadata**

- Fixes #[Link to Issue]

### DIFF
--- a/proxyd/server.go
+++ b/proxyd/server.go
@@ -641,7 +641,7 @@ func (s *Server) handleBatchRPC(ctx context.Context, reqs []json.RawMessage, isL
 				log.Error(
 					"error forwarding RPC batch",
 					"batch_size", len(elems),
-					"backend_group", group,
+					"backend_group", group.backendGroup,
 					"req_id", GetReqID(ctx),
 					"err", err,
 				)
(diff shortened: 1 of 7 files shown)


proxyd/server.go
```diff
@@ -641,7 +641,7 @@ func (s *Server) handleBatchRPC(ctx context.Context, reqs []json.RawMessage, isL
 				log.Error(
 					"error forwarding RPC batch",
 					"batch_size", len(elems),
-					"backend_group", group,
+					"backend_group", group.backendGroup,
```
dpulitano, line 644: logging fix which helps with monitoring

zhwrd (approved):

PR #438, opened 2025-07-28
feat(op-acceptor): parallel testing.

### LINKED ISSUE #184: [nat] Parallel tests
For performance, allow tests to be run in parallel.

scharissis:
## Description
Adds the ability to run multiple packages concurrently and makes this the new default.
The concurrency factor (number of packages to test concurrently) is determined automatically based on the system cores. It can also be manually specified instead with the new flag `--concurrency <N>`.
Also, the concurrent behaviour can be disabled with the new flag `--serial`.

## Tests
Testing shows significant speedups.

Example: running ALL existing sysgo/in-memory tests in the monorepo under directory `op-acceptance-tests/tests`:

BEFORE: 1h 42m
concurrency_before

AFTER: 14m
concurrency_after

## Metadata
- Fixes https://github.com/ethereum-optimism/infra/issues/184

### DIFF
--- a/op-acceptor/runner/runner.go
+++ b/op-acceptor/runner/runner.go
@@ -22,6 +22,9 @@ import (
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/trace"
 
+	"math"
+	"runtime"
+
 	"github.com/ethereum-optimism/infra/op-acceptor/flags"
 	"github.com/ethereum-optimism/infra/op-acceptor/logging"
 	"github.com/ethereum-optimism/infra/op-acceptor/metrics"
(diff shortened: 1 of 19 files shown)


op-acceptor/runner/runner.go
```diff
@@ -189,25 +202,87 @@ func (r *runner) RunAllTests(ctx context.Context) (*RunnerResult, error) {
...
+	result.WallClockTime = wallClockTime
+	result.Duration = wallClockTime // In serial mode, these are the same
 	result.Status = determineRunnerStatus(result)
 	result.Stats.EndTime = time.Now()
 	result.RunID = r.runID
 	return result, nil
 }
 
+// runAllTestsParallel implements the new parallel test execution logic
+func (r *runner) runAllTestsParallel(ctx context.Context) (*RunnerResult, error) {
```
serpixel, line 237: I think we should consider sharding the runs / running only affected tests if possible as a next step

op-acceptor/runner/runner.go
```diff
@@ -1577,3 +1652,101 @@ func (r *runner) parseTestOutputWithTimeout(output []byte, metadata types.Valida
...
+			"requested", requestedConcurrency,
+			"effective", effectiveConcurrency,
+			"workItems", numWorkItems)
+		return effectiveConcurrency
+	}
+
+	numCPU := runtime.NumCPU()
+	baseConcurrency := numCPU
+
+	var targetConcurrency int
```
serpixel, line 1715: I think in the long run we want this to be way more clever, specially as systems have efficiency and performance cores which do not provide the same output.

serpixel (approved):
🔥 Left a couple of comments for future improvements

PR #439, opened 2025-07-29
fix(op-acceptor): tests; fix flaky test

scharissis:


op-acceptor/runner/logging_test.go
```diff
@@ -306,32 +322,41 @@ gates:
...
 	}
 
-	time.Sleep(10 * time.Millisecond)
-
-	// Track which message we're expecting next
-	nextExpectedIndex := 0
-	timeout := time.After(1000 * time.Millisecond)
+	// Wait for all expected log messages with a reasonable timeout
+	receivedCount := 0
+	timeout := time.After(3 * time.Second) // More generous timeout for CI environments
```
serpixel: Would be great to have this based off an env variable that indicates we're on CI

serpixel (approved):

op-acceptor/runner/logging_test.go
ajsutton: 3 seconds is still a very tight timeout. CI servers can be very busy and it's not uncommon for a thread to not be scheduled for many seconds.  If this is just a timeout does it matter if it's longer than necessary? We should be able to continue once the expected messages are received so the timeout is only reached in cases of failure.

scharissis: Thank you both.
I've pushed a [change](https://github.com/ethereum-optimism/infra/pull/439/commits/92f7ebad20796ec90a481e96b6831ff4c742f698#diff-83e957ff7f83a137f0e4e26941541e2bacdf28585cd27c2d8da8ef122ee80766R325-R332) which I believe should address both of your feedbacks.

PR #447, opened 2025-08-20
chore(op-acceptor): clean; remove unused code identified by static analysis tools

scharissis:
- Used deadcode and staticcheck to identify 11 unused functions
- Removed ~600 lines of code across multiple packages


op-acceptor/cmd/main_test.go
```diff
@@ -202,7 +202,7 @@ gates:
 			// Create temporary test directory
 			tempDir, err := os.MkdirTemp("/tmp", "op-acceptor-test-")
 			require.NoError(t, err, "Failed to create temporary directory")
-			defer os.RemoveAll(tempDir)
+			defer func() { _ = os.RemoveAll(tempDir) }()
```
janjakubnanista, line 206: Why this change? Not that I am a go expert but this seems like just introducing a new lambda

op-acceptor/logging/filelogger_test.go
```diff
@@ -714,93 +714,6 @@ func TestHTMLSummarySink_WithSubtestsAndNetworkInfo(t *testing.T) {
 // TestExtractErrorInfoFromJSON verifies that error information is correctly extracted from test output
 func TestExtractErrorInfoFromJSON(t *testing.T) {
```
janjakubnanista: This seems like it left an empty test function behind

janjakubnanista: Oh and the `extractErrorData` is still there - we sure we want to remove the tests?

op-acceptor/cmd/main_test.go
scharissis: To appease the linter. It was complaining that we weren't checking the error code which [os.RemoveAll](https://pkg.go.dev/os#RemoveAll) returns.

op-acceptor/logging/filelogger_test.go
scharissis: Good spot, it's indeed still used, so I'll restore this test. Cheers

janjakubnanista (approved):

PR #448, opened 2025-08-21
refactor(op-acceptor): test runner

scharissis:
Refactors the monolithic 1,755-line runner/runner.go file into 5 focused components to improve maintainability and separation of concerns:

  - TestExecutor: Individual test execution and process management
  - OutputParser: Go test JSON output parsing with timeout handling
  - ResultCollector: Hierarchical result aggregation (gates/suites/tests)
  - TestCoordinator: Test workflow orchestration
  - JSONStore: Raw JSON storage and retrieval

### DIFF
--- a/op-acceptor/runner/adapters.go
+++ b/op-acceptor/runner/adapters.go
@@ -50,12 +50,6 @@ func (a *parallelRunnerAdapter) RunParallel(ctx context.Context, validators []ty
 	return nil
 }
 
-// SetConcurrency sets the concurrency level (this is handled at executor creation time)
-func (a *parallelRunnerAdapter) SetConcurrency(concurrency int) {
-	// The existing ParallelExecutor doesn't support runtime concurrency changes
-	// This would need to be set during creation in NewParallelExecutor
-}
-
 // GetConcurrency returns the current concurrency level
 func (a *parallelRunnerAdapter) GetConcurrency() int {
 	return a.executor.concurrency
--- a/op-acceptor/runner/adapters.go
+++ b/op-acceptor/runner/adapters.go
@@ -1,0 +1,106 @@
+package runner
+
+import (
+	"context"
+
+	"github.com/ethereum-optimism/infra/op-acceptor/types"
+)
+
+// parallelRunnerAdapter adapts the existing ParallelExecutor to the ParallelRunner interface
+type parallelRunnerAdapter struct {
+	executor *ParallelExecutor
+}
+
+// NewParallelRunnerAdapter creates an adapter for the existing ParallelExecutor
+func NewParallelRunnerAdapter(executor *ParallelExecutor) ParallelRunner {
+	if executor == nil {
+		return nil
+	}
+	return &parallelRunnerAdapter{
+		executor: executor,
+	}
+}
+
+// RunParallel adapts the existing ExecuteTests method to the interface
+func (a *parallelRunnerAdapter) RunParallel(ctx context.Context, validators []types.ValidatorMetadata, result *RunnerResult) error {
+	// Convert validators to work items
+	var workItems []TestWork
+	for _, validator := range validators {
+		gateName := validator.Gate
+		if gateName == "" {
+			gateName = "default"
+		}
+
+		workItems = append(workItems, TestWork{
+			Validator: validator,
+			GateID:    gateName,
+			SuiteID:   validator.Suite,
+			ResultKey: getTestKeyFromValidator(validator),
+		})
+	}
+
+	// Execute using the existing parallel executor
+	executorResult, err := a.executor.ExecuteTests(ctx, workItems)
+	if err != nil {
+		return err
+	}
+
+	// Copy results from executor result to the provided result
+	copyResults(executorResult, result)
+	return nil
+}
+
+// SetConcurrency sets the concurrency level (this is handled at executor creation time)
+func (a *parallelRunnerAdapter) SetConcurrency(concurrency int) {
+	// The existing ParallelExecutor doesn't support runtime concurrency changes
+	// This would need to be set during creation in NewParallelExecutor
+}
+
+// GetConcurrency returns the current concurrency level
+func (a *parallelRunnerAdapter) GetConcurrency() int {
+	return a.executor.concurrency
+}
+
+// noOpProgressIndicator provides a no-op implementation of ProgressIndicator
+type noOpProgressIndicator struct{}
+
+// NewNoOpProgressIndicator creates a progress indicator that does nothing
+func NewNoOpProgressIndicator() ProgressIndicator {
+	return &noOpProgressIndicator{}
+}
+
+func (n *noOpProgressIndicator) StartGate(gateName string, totalTests int)           {}
+func (n *noOpProgressIndicator) StartSuite(suiteName string, totalTests int)         {}
+func (n *noOpProgressIndicator) UpdateTest(testName string, status types.TestStatus) {}
+func (n *noOpProgressIndicator) CompleteSuite(suiteName string)                      {}
+func (n *noOpProgressIndicator) CompleteGate(gateName string)                        {}
+
+// Helper functions
+
+func getTestKeyFromValidator(validator types.ValidatorMetadata) string {
+	if validator.FuncName != "" {
+		return validator.Package + "::" + validator.FuncName
+	}
+	return validator.Package
+}
+
+func copyResults(from *RunnerResult, to *RunnerResult) {
+	if from == nil || to == nil {
+		return
+	}
+
+	// Copy basic fields
+	to.Status = from.Status
+	to.Duration = from.Duration
+	to.WallClockTime = from.WallClockTime
+	to.Stats = from.Stats
+
+	// Copy gates
+	if to.Gates == nil {
+		to.Gates = make(map[string]*GateResult)
+	}
+
+	for gateID, gate := range from.Gates {
+		to.Gates[gateID] = gate
+	}
+}
--- a/op-acceptor/runner/collector.go
+++ b/op-acceptor/runner/collector.go
@@ -6,6 +6,8 @@ import (
 	"github.com/ethereum-optimism/infra/op-acceptor/types"
 )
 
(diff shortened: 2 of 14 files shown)


op-acceptor/runner/adapters.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+// RunParallel adapts the existing ExecuteTests method to the interface
+func (a *parallelRunnerAdapter) RunParallel(ctx context.Context, validators []types.ValidatorMetadata, result *RunnerResult) error {
+	// Convert validators to work items
+	var workItems []TestWork
+	for _, validator := range validators {
+		gateName := validator.Gate
+		if gateName == "" {
+			gateName = "default"
```
serpixel, line 31: Is it possible that we will not have a gate defined? Is it for in-memory tests?

op-acceptor/runner/collector.go
```diff
@@ -0,0 +1,231 @@
...
+	}
+	// Get or create gate
+	gate, exists := result.Gates[gateName]
+	if !exists {
+		gate = &GateResult{
+			ID:          gateName,
+			Description: gateName,
+			Tests:       make(map[string]*types.TestResult),
+			Suites:      make(map[string]*SuiteResult),
+			Status:      types.TestStatusPass,
```
serpixel: Why are we adding by default the test result as passing?

op-acceptor/runner/collector.go
```diff
@@ -0,0 +1,231 @@
...
+		}
+		if gate.Status == types.TestStatusFail {
+			anyFailed = true
+		}
+	}
+
+	return c.determineStatusFromFlags(allSkipped, anyFailed)
+}
+
+func (c *resultCollector) determineStatusFromFlags(allSkipped, anyFailed bool) types.TestStatus {
```
serpixel: I am wondering if we require these methods to be fair, seems like we could simplify this

op-acceptor/runner/parser.go
```diff
@@ -0,0 +1,245 @@
...
+	// Set error message if test failed
+	if result.Status == types.TestStatusFail && errorMsg.Len() > 0 {
+		result.Error = fmt.Errorf("%s", strings.TrimSpace(errorMsg.String()))
+	}
+
+	// Handle skip status
+	if hasSkip && result.Status != types.TestStatusFail {
+		result.Status = types.TestStatusSkip
+		if errorMsg.Len() > 0 && result.Error == nil {
+			result.Error = fmt.Errorf("SKIP: %s", strings.TrimSpace(errorMsg.String()))
```
serpixel: Why are we labelling a skip as an error? Perhaps we just should add it as an info message

op-acceptor/runner/adapters.go
scharissis: Yes.
There's a new "ageless" mode

op-acceptor/runner/collector.go
scharissis: Optimism :)

For simplicity, I didn't want to introduce an 'Unknown' status, which meant this default either had to be Pass,Skip,Fail.
It then gets re-calculated later in `determineSuiteStatus()` and `determineGateStatus()`.

In the interest of time, I'm going to assume that you're concerned that it may result in false positives. I've changed the default to FAIL and added some tests, including to cover the edge case in which we have no tests in a gate.

op-acceptor/runner/parser.go
scharissis: Good spot, this was a mistake. I've corrected this.

op-acceptor/runner/collector.go
scharissis: I'm not sure what you mean by these methods being "fair".

I've gone ahead and simplified things by reusing the one package-level implementation.

serpixel (approved):

op-acceptor/runner/adapters.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+
+	// Copy results from executor result to the provided result
+	copyResults(executorResult, result)
+	return nil
+}
+
+// SetConcurrency sets the concurrency level (this is handled at executor creation time)
+func (a *parallelRunnerAdapter) SetConcurrency(concurrency int) {
+	// The existing ParallelExecutor doesn't support runtime concurrency changes
```
janjakubnanista: Maybe `panic` or at least log here?

op-acceptor/runner/collector.go
```diff
@@ -0,0 +1,233 @@
...
+
+	// Update statistics for a test result
+	UpdateStats(result *RunnerResult, gate *GateResult, suite *SuiteResult, test *types.TestResult)
+
+	// Finalize results and calculate statuses
+	FinalizeResults(result *RunnerResult)
+}
+
+// resultCollector implements ResultCollector
+type resultCollector struct{}
```
janjakubnanista, line 27: I'd add a line here to verify that this really implements `ResultsCollector`:

```go
var _ ResultCollector = (*ResultCollector)(nil)
```

op-acceptor/runner/executor.go
```diff
@@ -0,0 +1,207 @@
...
+type TestExecutor interface {
+	// Execute runs a single test or package based on the metadata provided.
+	// If metadata.FuncName is empty, it will execute all tests in the package.
+	Execute(ctx context.Context, metadata types.ValidatorMetadata) (*types.TestResult, error)
+
+	// ExecutePackage runs all tests in a specific package.
+	ExecutePackage(ctx context.Context, metadata types.ValidatorMetadata) (*types.TestResult, error)
+}
+
+// testExecutor implements TestExecutor
```
janjakubnanista, line 29: Same as above - add that line that confirms it really implements it

op-acceptor/runner/executor.go
```diff
@@ -0,0 +1,207 @@
...
+}
+
+// NewTestExecutor creates a new test executor
+func NewTestExecutor(testDir string, timeout time.Duration, goBinary string, envProvider func() Env,
+	cmdBuilder func(ctx context.Context, name string, arg ...string) (*exec.Cmd, func()),
+	outputParser OutputParser, jsonStore JSONStore) TestExecutor {
+
+	// Input validation
+	if testDir == "" {
+		panic("testDir cannot be empty")
```
janjakubnanista: Are we sure we want to `panic` here? Is it too difficult to return and handle errors instead?

op-acceptor/runner/jsonstore.go
```diff
@@ -0,0 +1,64 @@
+package runner
+
+import (
+	"sync"
+
+	"github.com/ethereum-optimism/infra/op-acceptor/logging"
+)
+
+// jsonStore implements JSONStore interface using FileLogger's RawJSONSink
+type jsonStore struct {
```
janjakubnanista, line 12: Same as above

op-acceptor/runner/parser.go
```diff
@@ -0,0 +1,244 @@
...
+	"encoding/json"
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/ethereum-optimism/infra/op-acceptor/types"
+)
+
+// outputParser implements OutputParser interface
+type outputParser struct{}
```
janjakubnanista, line 17: Same as above

op-acceptor/runner/parser.go
```diff
@@ -0,0 +1,244 @@
...
+	for scanner.Scan() {
+		line := scanner.Bytes()
+		event, err := parseTestEvent(line)
+		if err != nil {
+			continue
+		}
+
+		if isMainTestEvent(event, metadata.FuncName) {
+			processMainTestEvent(event, result, &testStart, &testEnd, &errorMsg, &hasSkip)
+		} else if event.Test != "" && (strings.Contains(event.Test, "/") || metadata.FuncName == "") {
```
janjakubnanista: I think it would add tons of readability if this `else` was also pulled out into a well named function, as it is I am struggling to understand what it means (as opposed to `isMainTestEvent` which is nice and understandable)

op-acceptor/runner/runner_test.go
```diff
@@ -1732,7 +1732,7 @@ gates:
 	require.Contains(t, result.SubTests, "TestSkipped", "Skipped test should be present in results")
 	skippedTest := result.SubTests["TestSkipped"]
 	require.Equal(t, types.TestStatusSkip, skippedTest.Status, "TestSkipped should have skip status")
-	require.Contains(t, skippedTest.Error.Error(), "This test is intentionally skipped", "Skip message should be preserved")
+	require.Nil(t, skippedTest.Error, "Skipped tests should not have error messages (skip reasons are not errors)")
```
janjakubnanista, line 1735: This seems like a big change - going from having error to not having it - are we sure this ain't gonna break some downstreams?

janjakubnanista (changes requested):
Some questions and small stuff, looks good otherwise!

janjakubnanista (approved):

op-acceptor/runner/adapters.go
scharissis: I've gone ahead and removed this entirely to simplify things.
Cheers.

op-acceptor/runner/executor.go
scharissis: Agreed. I've refactored this to be error-based.
Cheers

op-acceptor/runner/jsonstore.go
scharissis: Done

op-acceptor/runner/parser.go
scharissis: Done

op-acceptor/runner/parser.go
scharissis: Agreed. Extracted this to a function `isSubTestEvent`.
Cheers

op-acceptor/runner/runner_test.go
scharissis: It's a change, but not a breaking change, as it's not exposed externally (to consumers).
Good call out, cheers!

PR #450, opened 2025-08-25
feat(op-acceptor): just lint

scharissis:
The just lint command stopped working with newer versions of golangci-lint. This patches our local linting to try match that of CI's.

### DIFF
--- a/op-acceptor/justfile
+++ b/op-acceptor/justfile
@@ -40,4 +40,17 @@ stop-monitoring:
     docker compose -f 'docker-compose.yml' down
 
 lint:
-    golangci-lint run -E bodyclose,asciicheck,misspell,errorlint --timeout 5m --fix ./...
+    #!/bin/bash
+    # Run goimports with auto-fix (CI includes this but as a formatter, not linter)
+    echo "Fixing goimports formatting..."
+    if goimports -l . | grep -q .; then \
+        echo " - Found goimports issues; fixing..."; \
+        goimports -w .; \
+    fi
+    # Run golangci-lint matching CI config (excluding unsupported goimports linter and -e flags)
+    echo "Running golangci-lint..."
+    if [ -f .golangci.yml ]; then \
+        golangci-lint run -c .golangci.yml -E sqlclosecheck,bodyclose,asciicheck,misspell,errorlint --timeout "5m0s" ./...; \
+    else \
+        golangci-lint run -E sqlclosecheck,bodyclose,asciicheck,misspell,errorlint --timeout "5m0s" ./...; \
+    fi


ajsutton:
You should be using the same version of golangci-lint as in CI because it's installed via mise.  I just ran `make lint-go` locally and it passed as expected.

ajsutton:
Oh this is for op-acceptor - why isn't it just using the top level lint target like everything else does?

ajsutton:
ah, infra repo. Don't mind me - I'm lost.

But I'd still recommend not adding all this complexity and aligning more with the monorepo's approach which is just a simple invocation.

op-acceptor/justfile
```diff
@@ -40,4 +40,17 @@ stop-monitoring:
     docker compose -f 'docker-compose.yml' down
 
 lint:
-    golangci-lint run -E bodyclose,asciicheck,misspell,errorlint --timeout 5m --fix ./...
+    #!/bin/bash
+    # Run goimports with auto-fix (CI includes this but as a formatter, not linter)
+    echo "Fixing goimports formatting..."
+    if goimports -l . | grep -q .; then \
+        echo " - Found goimports issues; fixing..."; \
```
janjakubnanista, line 47: Is the goimports slow? If not i'd just skip the check and run this always

op-acceptor/justfile
```diff
@@ -40,4 +40,17 @@ stop-monitoring:
...
+        goimports -w .; \
+    fi
+    # Run golangci-lint matching CI config (excluding unsupported goimports linter and -e flags)
+    echo "Running golangci-lint..."
+    if [ -f .golangci.yml ]; then \
+        golangci-lint run -c .golangci.yml -E sqlclosecheck,bodyclose,asciicheck,misspell,errorlint --timeout "5m0s" ./...; \
+    else \
+        golangci-lint run -E sqlclosecheck,bodyclose,asciicheck,misspell,errorlint --timeout "5m0s" ./...; \
```
janjakubnanista, line 55: Nothing required: you can dedupe this line by creating a private recipe and pass -c in one case and -E in another. Private recipes start with `_` (just readme has some examples and guidance)

janjakubnanista (approved):

scharissis:
> ah, infra repo. Don't mind me - I'm lost.
> 
> But I'd still recommend not adding all this complexity and aligning more with the monorepo's approach which is just a simple invocation.

I like this approach, but it isn't so simple in this repo. CI is using an old version (1.65.4). Most folks will locally install the (only available) latest via `brew` which is 2.4.0 and has breaking changes. So the same command can't "just work" across both.

What I could do is modify all just/make commands and the CI to use a newer version.

ajsutton:
Or just start using mise and it will sync the version for you?

scharissis:
> Or just start using mise and it will sync the version for you?

Yeah, or that.
I did consider it, but it'll be a larger effort and also `mise` isn't particularly popular.
I'll re-investigate this avenue and try solve it for the entire repo (all projects).

scharissis:
I went ahead and introduced `mise` to this repository. It wasn't a small task, but I think it'll be better long-term.
https://github.com/ethereum-optimism/infra/pull/452

Thanks @ajsutton

PR #457, opened 2025-09-08
op-acceptor: add console progress indicator

### LINKED ISSUE #456: [acceptance] logging - improve a progress indicator
Our [progress indicator](https://github.com/ethereum-optimism/infra/blob/main/op-acceptor/runner/runner.go#L232-L233) is currently a placeholder; let's implement something (even if it's simple) to improve the UX around awareness of test, package and overall run progress.

bitwiseguy:
# Overview
Users of `op-acceptor` would like more information to help debug tests that are stalled or taking longer than expected. Prior to this pr there was the framework for a `ProgressIndicator` to provide feedback to the user about testing progress, but only a no-op indicator was used. This pr adds a `ConsoleProgressIndicator`, which will print logs to the console while tests are running. Logs have this format:
```
INFO [09-11|14:12:31.263] progress update                          gate=gateless suite= completed=1 total=32 percent=3.1% numRunning=16 longestRunning="./tests/interop/message (5s), ./tests/interop/prep (5s), ./tests/interop/proofs (5s), +13 more"
```

# Details
Currently, the `ConsoleProgressIndicator` is disabled by default. To enable it, set the `ShowProgress = true` field on the `op-acceptor` config. There is also a new config field for `ProgressInterval` (how often to print a log/update), which defaults to 30s.

Only the 3 longest running tests are logged at any given time (to keep logging printout minimal while still providing the most useful information). As tests complete, they are removed from the data structures that are used to populate the progress logs.

# Tests
Added `progress_test.go` to test the new functionality by capturing logs and parsing them for expected output.

## Related Issues
* Closes https://github.com/ethereum-optimism/infra/issues/456

### DIFF
--- a/op-acceptor/runner/adapters.go
+++ b/op-acceptor/runner/adapters.go
@@ -55,20 +55,6 @@ func (a *parallelRunnerAdapter) GetConcurrency() int {
 	return a.executor.concurrency
 }
 
-// noOpProgressIndicator provides a no-op implementation of ProgressIndicator
-type noOpProgressIndicator struct{}
-
-// NewNoOpProgressIndicator creates a progress indicator that does nothing
-func NewNoOpProgressIndicator() ProgressIndicator {
-	return &noOpProgressIndicator{}
-}
-
-func (n *noOpProgressIndicator) StartGate(gateName string, totalTests int)           {}
-func (n *noOpProgressIndicator) StartSuite(suiteName string, totalTests int)         {}
-func (n *noOpProgressIndicator) UpdateTest(testName string, status types.TestStatus) {}
-func (n *noOpProgressIndicator) CompleteSuite(suiteName string)                      {}
-func (n *noOpProgressIndicator) CompleteGate(gateName string)                        {}
-
 // Helper functions
 
 func getTestKeyFromValidator(validator types.ValidatorMetadata) string {
--- a/op-acceptor/runner/coordinator.go
+++ b/op-acceptor/runner/coordinator.go
@@ -15,15 +15,6 @@ type ParallelRunner interface {
 	GetConcurrency() int
 }
 
-// ProgressIndicator interface for UI updates
-type ProgressIndicator interface {
-	StartGate(gateName string, totalTests int)
-	StartSuite(suiteName string, totalTests int)
-	UpdateTest(testName string, status types.TestStatus)
-	CompleteSuite(suiteName string)
-	CompleteGate(gateName string)
-}
-
 // TestCoordinator orchestrates test execution
 type TestCoordinator interface {
 	// Run all tests with the given validators
@@ -31,6 +22,7 @@ type TestCoordinator interface {
 
 	// Run tests for a specific gate
 	RunGate(ctx context.Context, gateName string, validators []types.ValidatorMetadata) (*GateResult, error)
+	GetUI() ProgressIndicator
 }
 
 // testCoordinator implements TestCoordinator
@@ -49,6 +41,11 @@ func NewTestCoordinator(executor TestExecutor, collector ResultCollector, parall
 		parallelRunner: parallelRunner,
 		ui:             ui,
 	}
+}
+
+// GetUI returns the progress indicator
+func (c *testCoordinator) GetUI() ProgressIndicator {
+	return c.ui
 }
 
 // Run orchestrates the execution of all tests
@@ -85,6 +82,11 @@ func (c *testCoordinator) RunGate(ctx context.Context, gateName string, validato
 		}
 
 		for _, validator := range suiteTests {
+			// Notify progress indicator that test is starting
+			if c.ui != nil {
+				c.ui.StartTest(validator.GetName())
+			}
+
 			testResult, err := c.executor.Execute(ctx, validator)
 			if err != nil {
 				log.Error("Test execution failed", "test", validator.GetName(), "error", err)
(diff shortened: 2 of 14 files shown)


op-acceptor/runner/adapters.go
```diff
@@ -55,20 +55,6 @@ func (a *parallelRunnerAdapter) GetConcurrency() int {
...
-func NewNoOpProgressIndicator() ProgressIndicator {
-	return &noOpProgressIndicator{}
-}
-
-func (n *noOpProgressIndicator) StartGate(gateName string, totalTests int)           {}
-func (n *noOpProgressIndicator) StartSuite(suiteName string, totalTests int)         {}
-func (n *noOpProgressIndicator) UpdateTest(testName string, status types.TestStatus) {}
-func (n *noOpProgressIndicator) CompleteSuite(suiteName string)                      {}
-func (n *noOpProgressIndicator) CompleteGate(gateName string)                        {}
-
```
bitwiseguy, line 71: Moved to the new `progress.go` file to help organize the code

op-acceptor/runner/coordinator.go
```diff
@@ -15,22 +15,14 @@ type ParallelRunner interface {
...
 
-// ProgressIndicator interface for UI updates
-type ProgressIndicator interface {
-	StartGate(gateName string, totalTests int)
-	StartSuite(suiteName string, totalTests int)
-	UpdateTest(testName string, status types.TestStatus)
-	CompleteSuite(suiteName string)
-	CompleteGate(gateName string)
-}
-
```
bitwiseguy, line 26: Moved to the new `progress.go` file to help organize code

op-acceptor/runner/parallel.go
```diff
@@ -31,10 +31,11 @@ type ParallelExecutor struct {
 	concurrency int
 	log         log.Logger
 	resultMgr   *ResultHierarchyManager
+	ui          ProgressIndicator
 }
 
 // NewParallelExecutor creates a new parallel test executor with validation
-func NewParallelExecutor(runner *runner, concurrency int) *ParallelExecutor {
+func NewParallelExecutor(runner *runner, concurrency int, ui ProgressIndicator) *ParallelExecutor {
```
bitwiseguy: We could probably avoid adding this new `ui` arg and instead grab the `ui` from the `runner.coordinator`. But sometimes the `runner.coordinator == nil` when `NewParallelExecutor` is called so we'd have to account for that or refactor other code to restrict when `NewParallelExecutor` is called. It feels redundant to set the `ui` on both the coordinator and this executor

op-acceptor/runner/progress.go
```diff
@@ -0,0 +1,222 @@
...
+
+// reportProgress logs the current progress status
+func (c *consoleProgressIndicator) reportProgress() {
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+
+	// Don't report if no gate is active
+	if c.currentGate == "" {
+		return
+	}
```
scharissis: Does this work with 'gateless mode'?

op-acceptor/runner/progress.go
```diff
@@ -0,0 +1,222 @@
...
+
+	// Build running tests string
```
scharissis: Have you considered making this string-building a helper function?
That would make it easier to unit test this logic too.

op-acceptor/runner/parallel.go
scharissis: I agree this would be better.
Shall we tackle it now or later in your opinion? For example, I could perhaps do that refactor/fix and then you could rebase this on top of that?

op-acceptor/runner/parallel.go
```diff
@@ -249,3 +266,33 @@ func (r *runner) collectTestWork() []TestWork {
...
+// initializeProgressTracking sets up data structures to concurrently
+// track progress for each gate and suite in the scheduled work items
+func (pe *ParallelExecutor) initializeProgressTracking(workItems []TestWork) {
+	pe.log.Info("Initializing parallel progress tracking")
+
+	// Group work items by gate
+	gateGroups := make(map[string][]TestWork)
+	for _, item := range workItems {
+		gateGroups[item.GateID] = append(gateGroups[item.GateID], item)
+	}
```
scharissis, line 303: Is this necessary? We have either 0 or 1 gates at any one time.

op-acceptor/runner/parallel.go
scharissis: I did the refactor here: https://github.com/ethereum-optimism/infra/pull/458

bitwiseguy: Merged your refactor pr into this branch

op-acceptor/runner/progress.go
bitwiseguy: Removed this early return. It should work fine without this check, the log will just have less info

op-acceptor/runner/parallel.go
bitwiseguy: If there's any chance we run multiple gates concurrently in the future this would be nice to leave in. Do we know for certain we'll only ever want to run one gate at a time?

op-acceptor/runner/progress.go
bitwiseguy: Good call. Split out the helper function and added a few unit tests for it

op-acceptor/runner/parallel.go
scharissis: It's not a certainty going forward; happy to leave this in.

scharissis (approved):

PR #464, opened 2025-09-15
proxyd: Add fusaka blob transaction support

mslipper:
Adds support for post-Fusaka blob transactions, which include cell proofs. As part of this PR, I updated the Geth and monorepo versions proxyd uses. In doing so, I found a bug where proxy would panic if pre-EIP-155 transactions were ingested. I fixed that.

Test vectors are from https://gist.github.com/protolambda/25a599a3ab05fbf7d938b21ca94f2e1c.


proxyd/server.go
```diff
@@ -831,7 +831,15 @@ func (s *Server) genericRateLimitSender(ctx context.Context, tx *types.Transacti
 		return txpool.ErrInvalidSender
 	}
 
-	signer := types.LatestSignerForChainID(tx.ChainId())
+	var signer types.Signer
+	// If you pass in a zero chain ID, types.LatestSignerForChainID panics. So we need to handle that case
+	// manually.
+	if tx.ChainId().Cmp(common.Big0) == 0 {
```
sebastianst:
```suggestion
	if tx.ChainId().Sign() == 0 {
```

proxyd/server.go
```diff
@@ -831,7 +831,15 @@ func (s *Server) genericRateLimitSender(ctx context.Context, tx *types.Transacti
...
+		signer = new(types.HomesteadSigner)
```
sebastianst, line 838: So this wouldn't be an error case? How could a tx for a zero chain id be valid?

mslipper: if it's pre-EIP 155. the chain ID is zero if it's not replay-protected.

zhwrd (approved):

zhwrd (approved):

scharissis:
LGTM

PR #474, opened 2025-09-23
fix(op-acceptor): improved logging

scharissis:
Various logging fixes and improvements:
- Fix parser to capture subtest stdout output
- Handle plain text output when JSON parsing produces no content
- Ensure HTML report links match actual created files
- Add file existence checking in LogPathGenerator
- Strip ANSI escape sequences from plaintext output
- Added tests for all the above

### DIFF
--- a/op-acceptor/logging/filelogger.go
+++ b/op-acceptor/logging/filelogger.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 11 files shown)


op-acceptor/logging/filelogger.go
```diff
@@ -22,6 +23,14 @@ const (
 	RunDirectoryPrefix  = "testrun-" // Standardized prefix for run directories
 )
 
+// ansiRegex matches ANSI escape sequences
+var ansiRegex = regexp.MustCompile(`\x1b\[[0-9;]*m`)
```
janjakubnanista, line 27: https://github.com/acarl005/stripansi/blob/master/stripansi.go this package has a more comprehensive ansi stripping

janjakubnanista (approved):

PR #475, opened 2025-09-23
feat(op-acceptor): flake shake

scharissis:
## Description
Introduces a new runner mode  - "flake-shake", which runs tests many times to identify flaky ones.
HTML and JSON reports are produced for CI artifacts. These will be used to aggregate results across multiple runners and , possibly, runs.

Screenshot 2025-09-24 at 09 35 19
Screenshot 2025-09-24 at 09 35 45

### DIFF
--- a/op-acceptor/runner/flake_shake.go
+++ b/op-acceptor/runner/flake_shake.go
@@ -1,0 +1,311 @@
+package runner
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"html/template"
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/ethereum-optimism/infra/op-acceptor/types"
+	"github.com/ethereum/go-ethereum/log"
+)
+
+// FlakeShakeResult represents aggregated results for a test across multiple runs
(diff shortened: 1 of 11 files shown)


op-acceptor/runner/flake_shake.go
```diff
@@ -0,0 +1,311 @@
...
+	results := make(map[string][]types.TestResult)
+
+	// Run tests multiple times
+	for i := 1; i <= f.iterations; i++ {
+		f.log.Info("Running iteration", "iteration", i, "total", f.iterations)
+
+		// Run all tests once
+		runResult, err := f.baseRunner.RunAllTests(ctx)
+		if err != nil {
+			f.log.Error("Failed to run tests", "iteration", i, "error", err)
```
serpixel, line 75: Will this still report the entire stack trace at the end? My concern is that we may lose failure information that may differ between runs.

serpixel (approved):
Just one comment

op-acceptor/runner/flake_shake.go
scharissis: It is partial (not all traces from all runs).
I'll think about how to best improve this for a follow-up PR.
Thanks!

PR #476, opened 2025-09-25
chore: remove unused min func

cargoedit:
**Description**

Go 1.21 adds built-in functions min and max.   https://tip.golang.org/doc/go1.21
Therefore, we can directly remove our own implementations and use the built-in function with the same names. 

**Tests**

**Additional context**

**Metadata**

- Fixes #[Link to Issue]


scharissis (approved):
Thanks

PR #485, opened 2025-10-05
feat(op-acceptor): skip gates.

scharissis:
Added the ability to skip certain gates.

### DIFF
--- a/op-acceptor/README.md
+++ b/op-acceptor/README.md
@@ -59,6 +59,26 @@ DEVNET_ENV_URL=kt://isthmus-devnet op-acceptor \
     --log.level INFO
 ```
 
+## Excluding tests via skip gates
+You can exclude tests that belong to one or more gates from every run.
+
+- **Flag**: `--exclude-gates` (comma-separated gate IDs). Default: no exclusions.
+- **Env var**: `ACCEPTOR_EXCLUDE_GATES` overrides the flag if set. Set to an empty string to disable all exclusions.
+- **Gateless note**: in gateless mode, exclusions are applied only if a validators YAML is provided (so gate membership can be resolved).
+
+Examples:
+
+```bash
+# No default exclusions; to exclude, specify one or more gates
+op-acceptor --validators acceptance-tests.yaml --gate base --exclude-gates "flake-shake"
+
+# Exclude multiple gates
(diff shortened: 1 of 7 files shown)


op-acceptor/README.md
```diff
@@ -59,6 +59,29 @@ DEVNET_ENV_URL=kt://isthmus-devnet op-acceptor \
...
+
+- **Flag**: `--exclude-gates` (comma-separated gate IDs). Default: when not set anywhere, tests in gate `flake-shake` are excluded by default.
+- **Env var**: `ACCEPTOR_EXCLUDE_GATES` overrides the flag if set. Set to an empty string to disable all exclusions.
+- **Gateless note**: in gateless mode, exclusions are applied only if a validators YAML is provided (so gate membership can be resolved).
+
+Examples:
+
+```bash
+# Default behavior skips tests listed under the `flake-shake` gate in validators YAML
+op-acceptor --validators acceptance-tests.yaml --gate base --testdir ./...
```
janjakubnanista: Out of curiosity: this seems like a very specific default value - why not exclude nothing by default?

op-acceptor/config.go
```diff
@@ -100,28 +103,65 @@ func NewConfig(ctx *cli.Context, log log.Logger, testDir string, validatorConfig
...
+		ExcludeGates:         excludeGates,
 	}, nil
 }
+
+// parseExcludeGates determines which gates to exclude based on env/flag.
+// If env var ACCEPTOR_EXCLUDE_GATES is set, it overrides the flag value (even if empty).
+// Default behavior: if both are unset/empty, default to ["flake-shake"].
+// Users can disable skipping entirely by explicitly setting an empty string.
+func parseExcludeGates(flagVal string, envVal string, envPresent bool) []string {
+	// If env var present, it overrides the flag even if empty
```
janjakubnanista: I think this might not be what you would expect from a CLI interface - usually a flag has higher priority than an env variable

janjakubnanista (approved):
Going to approve but check my comments out, code is fine but the UX seems very non-standard

op-acceptor/README.md
scharissis: I was unsure about this myself. Given your comment, I've changed it back to an empty default. Thanks!

op-acceptor/config.go
scharissis: Agreed; changed it back to the usual convention. Cheers

PR #494, opened 2025-10-22
feat(proxyd): max block range config

charlesma4:
**Description**

Max block range config to limit block range for potentially expensive rpc requests like `eth_getLogs` and `eth_newFilter`. The config applies across all routing strategies, and will only be applied if the config value is set. Otherwise, it reverts to original behavior, with limits for `consensus_aware` backends only if `consensus_max_block_range` is set.

Note, if `max_block_range` is set, the above two requests will not accept block tags, to prevent unbounded queries.

Max block range will override `ConsensusMaxBlockRange` if set.

**Tests**

Integration test for `eth_getLogs/eth_newFilter` to test that the config value is working as intended.

**Additional context**

This is from a Spearbit audit in Cantina that Unichain recently did.

```
Description
Some operations like eth_getLogs can take quite a while, especially if a user requests abnormally large block ranges.

This can be problematic when the proxyd service considers a timeout of 5 seconds for any requests that it has forwarded. It will mark down an error case if any backend takes longer than 5 seconds to service such a request. A backend with too many errors will eventually be considered an "unhealthy" backend and proxyd will stop using that backend.

A user can intentionally send many requests that result in a timeout. This is the cae because the nodes used in the Unichain deployment do not have limits configured needed to ensure that nodes are able to guarantee the 5 second latency contract that proxyd expects.

Executing this approach on a large scale can cause proxyd to consider an entire backend group as unhealthy, and stop servicing (some) legitimate requests.

Recommendation
Execute a range of simple benchmarks that can be used to inform an initial set of limits for the rpc calls that are prone to long execution times.

An alternative approach is to leverage monitoring of the RPC nodes and basing the limits on historical observations.

It might also be possible to leverage the existing monitoring stack to detect attacks attempting to exploit this flaw.
```

**Metadata**

- Fixes #[Link to Issue]

### DIFF
--- a/proxyd/backend.go
+++ b/proxyd/backend.go
@@ -1708,6 +1708,7 @@ func (bg *BackendGroup) OverwriteConsensusResponses(rpcReqs []*RPCReq, overridde
 		safe:          bg.Consensus.GetSafeBlockNumber(),
 		finalized:     bg.Consensus.GetFinalizedBlockNumber(),
 		maxBlockRange: bg.Consensus.maxBlockRange,
+		consensusMode: true,
 	}
 
 	for i, req := range rpcReqs {
@@ -1748,6 +1749,7 @@ func (bg *BackendGroup) OverwriteNonConsensusRequests(rpcReqs []*RPCReq, overrid
 		safe:          0,
 		finalized:     0,
 		maxBlockRange: bg.maxBlockRange,
+		consensusMode: false,
 	}
 
 	rewrittenReqs := make([]*RPCReq, 0, len(rpcReqs))
--- a/proxyd/backend.go
+++ b/proxyd/backend.go
@@ -926,6 +926,7 @@ type BackendGroup struct {
 	FallbackBackends       map[string]bool
 	routingStrategy        RoutingStrategy
 	multicallRPCErrorCheck bool
+	maxBlockRange          uint64
 }
 
 func (bg *BackendGroup) GetRoutingStrategy() RoutingStrategy {
@@ -969,6 +970,8 @@ func (bg *BackendGroup) Forward(ctx context.Context, rpcReqs []*RPCReq, isBatch 
 	// We also rewrite block tags to enforce compliance with consensus
 	if bg.Consensus != nil {
 		rpcReqs, overriddenResponses = bg.OverwriteConsensusResponses(rpcReqs, overriddenResponses, rewrittenReqs)
+	} else if bg.maxBlockRange > 0 {
+		rpcReqs, overriddenResponses = bg.OverwriteNonConsensusRequests(rpcReqs, overriddenResponses)
 	}
 
 	rpcRequestsTotal.Inc()
@@ -1725,6 +1728,46 @@ func (bg *BackendGroup) OverwriteConsensusResponses(rpcReqs []*RPCReq, overridde
 				)
 			} else {
 				res.Error = ErrParseErr
+			}
+		case RewriteOverrideResponse:
+			overriddenResponses = append(overriddenResponses, &indexedReqRes{
+				index: i,
+				req:   req,
+				res:   &res,
+			})
+		case RewriteOverrideRequest, RewriteNone:
+			rewrittenReqs = append(rewrittenReqs, req)
+		}
+	}
+	return rewrittenReqs, overriddenResponses
+}
+
+func (bg *BackendGroup) OverwriteNonConsensusRequests(rpcReqs []*RPCReq, overriddenResponses []*indexedReqRes) ([]*RPCReq, []*indexedReqRes) {
+	rctx := RewriteContext{
+		latest:        0,
+		safe:          0,
+		finalized:     0,
+		maxBlockRange: bg.maxBlockRange,
+	}
+
+	rewrittenReqs := make([]*RPCReq, 0, len(rpcReqs))
+
+	for i, req := range rpcReqs {
+		res := RPCRes{JSONRPC: JSONRPCVersion, ID: req.ID}
+		result, err := RewriteTags(rctx, req, &res)
+		switch result {
+		case RewriteOverrideError:
+			overriddenResponses = append(overriddenResponses, &indexedReqRes{
+				index: i,
+				req:   req,
+				res:   &res,
+			})
+			if errors.Is(err, ErrRewriteRangeTooLarge) {
+				res.Error = ErrInvalidParams(
+					fmt.Sprintf("block range greater than %d max", rctx.maxBlockRange),
+				)
+			} else {
+				res.Error = ErrInvalidParams(err.Error())
 			}
 		case RewriteOverrideResponse:
 			overriddenResponses = append(overriddenResponses, &indexedReqRes{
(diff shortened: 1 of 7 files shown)


proxyd/backend.go
```diff
@@ -969,6 +971,8 @@ func (bg *BackendGroup) Forward(ctx context.Context, rpcReqs []*RPCReq, isBatch
 	// We also rewrite block tags to enforce compliance with consensus
 	if bg.Consensus != nil {
 		rpcReqs, overriddenResponses = bg.OverwriteConsensusResponses(rpcReqs, overriddenResponses, rewrittenReqs)
+	} else {
+		rpcReqs, overriddenResponses = bg.ValidateBlockRangeLimits(rpcReqs, overriddenResponses)
 	}
```
jelias2: If the proxyd is using consensus based routing the validateBlockRange will use the `consensus_max_block_range` configuration.


In the PR its referenced, 
```
The config applies across all routing strategies, and will only be applied if the config value is set. Otherwise, it reverts to original behavior, with no limits.
```
Can you confirm the desired behavior. It looks like it only will work with non-consensus backend groups at this point in time

proxyd/backend.go
```diff
@@ -1738,3 +1742,128 @@ func (bg *BackendGroup) OverwriteConsensusResponses(rpcReqs []*RPCReq, overridde
...
+		if req.Method != "eth_getLogs" && req.Method != "eth_newFilter" {
+			validReqs = append(validReqs, req)
+			continue
+		}
+
+		var params []map[string]interface{}
+		if err := json.Unmarshal(req.Params, &params); err != nil || len(params) == 0 {
+			validReqs = append(validReqs, req)
+			continue
+		}
```
jelias2: Hmm can this code be explained a bit more?

It looks like if the json unmarshalling fails, or if the parameters are empty, the code is acknowledging that the request is valid by adding it to valid requests?

If the request can't be parsed correctly should it be forwarded to the backend? Probably not

proxyd/integration_tests/block_range_limit_test.go
```diff
@@ -0,0 +1,140 @@
...
+			expectError:  true,
+			errorMessage: "block tags (latest/pending/safe/finalized) are not allowed",
+		},
+		{
+			name:         "eth_getLogs with finalized tag (rejected)",
+			method:       "eth_getLogs",
+			params:       `[{"fromBlock":"0x0","toBlock":"finalized"}]`,
+			expectError:  true,
+			errorMessage: "block tags (latest/pending/safe/finalized) are not allowed",
+		},
```
jelias2: Can a couple more test cases be added
1. Batched RPC requests with both eth_gethLogs and eth_getFilter
2. Test cases for parseBlockParam errors being returned to the client

proxyd/backend.go
```diff
@@ -1738,3 +1742,128 @@ func (bg *BackendGroup) OverwriteConsensusResponses(rpcReqs []*RPCReq, overridde
...
+			})
+		} else {
+			validReqs = append(validReqs, req)
+		}
+	}
+
+	return validReqs, overriddenResponses
+}
+
+func parseBlockParam(raw interface{}) (uint64, error) {
```
jelias2: There already exists lots of logic within rewriter.go that handles rewriting requests. I'd encourage look at that package and seeing if there is code that could be reused and avoid de-duplication from this PR, or placing specific rewrite log in rewriter.go so that is can be reused. 

Could something like RewriteRequest be used within this overall function? Possibly modify the rewriter package to support the block range constraints for this functionality?

proxyd/backend.go
charlesma4: yes will correct this, if disabled then only consensus aware backends will have a max block range

proxyd/backend.go
charlesma4: was initially thinking that if we can't validate for whatever reason, we just err on the side of caution and fail-open

but on second thought I agree, I don't see any cases where a failure here would mean it's a valid request

maybe we should unmarshal in the caller function, and pass the params in? then the params can be used across other functions that may need to unmarshal as well, and marking a request as invalid would make more sense there too, as opposed to doing it in this block range validation function

charlesma4: seems this unmarshal attempt is pretty infrequent, going to leave it in this function but fail-close as suggested

proxyd/proxyd.go
```diff
@@ -298,13 +298,19 @@ func Start(config *Config) (*Server, func(), error) {
 				)
 		}
 
+		maxBlockRange := bg.ConsensusMaxBlockRange
+		if bg.MaxBlockRange > 0 {
+			maxBlockRange = bg.MaxBlockRange
+		}
+
```
jelias2, line 306: If the code is using consensus mode, MaxBlockRange will override ConsensusMaxBlockRange. I think consensus Max Block range should still be respected. Would suggest something like
```
maxBlockRange := uint64(0)
if bg.Consensus != nil {
    // When consensus is enabled, use ConsensusMaxBlockRange
    maxBlockRange = bg.ConsensusMaxBlockRange
} else {
    // When consensus is disabled, use MaxBlockRange
    maxBlockRange = bg.MaxBlockRange
}
```

proxyd/proxyd.go
```diff
@@ -511,8 +517,8 @@ func Start(config *Config) (*Server, func(), error) {
 			if bgcfg.ConsensusMinPeerCount > 0 {
 				copts = append(copts, WithMinPeerCount(uint64(bgcfg.ConsensusMinPeerCount)))
 			}
-			if bgcfg.ConsensusMaxBlockRange > 0 {
-				copts = append(copts, WithMaxBlockRange(bgcfg.ConsensusMaxBlockRange))
+			if bg.maxBlockRange > 0 {
+				copts = append(copts, WithMaxBlockRange(bg.maxBlockRange))
 			}
```
jelias2, line 523: This looks like bg.MaxBlockRange is overriding the consensusMaxBlockRange field. This doesn't look like the correct behavior? Should MaxBlockRange and ConsensusBlockRange be separate from each other?

proxyd/rewriter.go
```diff
@@ -153,13 +153,31 @@ func rewriteRange(rctx RewriteContext, req *RPCReq, res *RPCRes, pos int) (Rewri
...
 
-	// if either fromBlock or toBlock is defined, default the other to "latest" if unset
 	_, hasFrom := p[pos]["fromBlock"]
 	_, hasTo := p[pos]["toBlock"]
-	if hasFrom && !hasTo {
-		p[pos]["toBlock"] = "latest"
-	} else if hasTo && !hasFrom {
-		p[pos]["fromBlock"] = "latest"
+
+	nonConsensusMode := rctx.latest == 0
```
jelias2: Rather than settting rctx.latest == 0 to be set the rewrite conext to non-consensus mode, could a boolean be added to the rewrite conext to be an explicit way to determine rewrite should be ConsensusMode

`ConsensusMode bool`

proxyd/proxyd.go
charlesma4: That was the original intention, to have maxBlockRange override the consensus config only if it's set.

Was thinking if we want customized block ranges depending on routing strategy, we should deprecate both consensus and normal maxBlockRange, and have some sort of maxBlockRangeMap

proxyd/proxyd.go
charlesma4: https://github.com/ethereum-optimism/infra/pull/494/files#r2469564019

jelias2 (approved):

PR #495, opened 2025-10-22
feat(proxyd): verify last XFF IP matches client IP

charlesma4:
**Description**

Adds config to enable XFF verification, which compares the last IP in the XFF address list against the remote address (client IP). This protects against IP spoofing, which attackers can use to exhaust rate limits of other IPs.

Proxyd will automatically disable this verification if the rate limit header is overridden to not be `x-forwarded-for`.

**Tests**

Unit tests added for all new code

**Additional context**

```
Sender Verification
It is possible that attackers might connect to proxyd directly in misconfigured deployments.

To ensure correct functioning proxyd should verify that the last entry in the x-forwarded-for chain is equal to the connection initiator. Without such a check the trust chain control would be less effective, as an attacker could simply mirror a request as if it has the expected trusted proxy IP addresses.
```

**Metadata**

- Fixes #[Link to Issue]

### DIFF
--- a/proxyd/backend.go
+++ b/proxyd/backend.go
@@ -1583,6 +1583,7 @@ func stripXFF(xff string) string {
 	return strings.TrimSpace(ipList[0])
 }
 
+// getLastXFF extracts the last IP from X-Forwarded-For chain (e.g., "1.2.3.4, 5.6.7.8" -> "5.6.7.8")
 func getLastXFF(xff string) string {
 	if xff == "" {
 		return ""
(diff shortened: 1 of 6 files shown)


yashvardhan-kukreja:
Hi, just doublechecking the intent behind checking the last value in XFF header considering the current nature of code where it extracts and uses the first value in xff header.

charlesma4:
> Hi, just doublechecking the intent behind checking the last value in XFF header considering the current nature of code where it extracts and uses the first value in xff header.

Yes that is the intended behavior, that the connection initiator matches the last address in XFF (as is expected).

However, realizing this isn't a great solution, because attackers can still spoof the first IP in the XFF list, provided they append their own IP in XFF. It's not a complete solution unless we have some list of trusted proxies, which would be difficult to maintain. Closing this as it's not a complete solution

PR #496, opened 2025-10-23
chore(op-acceptor): readme; path semantics

scharissis:


joshklop (approved):
Thanks! I haven't tested this but I trust it works as documented :)

janjakubnanista (approved):

PR #497, opened 2025-10-27
chore: Switch to blobless checkout

janjakubnanista:
**Description**

CircleCI will [switch all checkouts to blobless on November 3rd](https://circleci.com/changelog/default-method-used-to-checkout-code-from-your-repository-is-changing-on-nov/). This change gets us there ahead of time so that we can check that everything works - if not, we'll switch to `full` checkout for the time being.


scharissis (approved):
LGTM

PR #500, opened 2025-10-30
clean(op-acceptor): subtest timeouts refactor

scharissis:
Refactored this code a bit as there was a divergent timeout handling.

* Changed timeout handling so completed sub-tests retain their real status, and only sub-tests that started but didn’t finish are marked as timed out; when start time is unknown, we now set the sub-test duration to the full timeout.
* The runner’s timeout path now delegates to the parser’s timeout logic and appends the actual duration to the error for clarity.
* Added tests to cover the preserved-completed-subtests behavior and verified the full test suite and linter pass.

### DIFF
--- a/op-acceptor/runner/parser.go
+++ b/op-acceptor/runner/parser.go
@@ -77,20 +77,103 @@ func (p *outputParser) Parse(output []byte, metadata types.ValidatorMetadata) *t
 
 // ParseWithTimeout parses test output for tests that exceeded timeout
 func (p *outputParser) ParseWithTimeout(output []byte, metadata types.ValidatorMetadata, timeout time.Duration) *types.TestResult {
-	result := p.Parse(output, metadata)
+	// If no output, return a minimal timeout result
+	if len(output) == 0 {
+		return &types.TestResult{
+			Metadata: metadata,
+			Status:   types.TestStatusFail,
+			Error:    fmt.Errorf("TIMEOUT: Test timed out after %v", timeout),
+			SubTests: make(map[string]*types.TestResult),
+			TimedOut: true,
+		}
+	}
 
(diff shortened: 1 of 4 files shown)


op-acceptor/runner/parser.go
```diff
@@ -77,20 +77,103 @@ func (p *outputParser) Parse(output []byte, metadata types.ValidatorMetadata) *t
...
+
+	subTestStatuses := make(map[string]types.TestStatus)
+	subTestStartTimes := make(map[string]time.Time)
+
+	scanner := bufio.NewScanner(bytes.NewReader(output))
+	for scanner.Scan() {
+		line := scanner.Bytes()
+		event, err := parseTestEvent(line)
+		if err != nil {
+			// Be lenient on JSON parsing during timeout scenarios
```
janjakubnanista, line 108: I'd smack a log line here just for us to be able to see whether we have maybe something we can improve in the parser or smth

op-acceptor/runner/parser.go
```diff
@@ -77,20 +77,103 @@ func (p *outputParser) Parse(output []byte, metadata types.ValidatorMetadata) *t
...
+			continue
+		}
+
+		// Attach any main test output to the error for context
+		if isMainTestEvent(event, metadata.FuncName) {
+			if event.Action == ActionOutput {
+				output := strings.TrimSpace(event.Output)
+				if output != "" && result.Error != nil {
```
janjakubnanista, line 116: Here I'd say: if we have some output and don't have any error, let's add some `unknown error` or smth, maybe the output is valuable.

Just having PTSD from weeks of debugging stuff with no output y'know 🤣

janjakubnanista (approved):

PR #525, opened 2025-12-06
fix(op-acceptor): Stream stdout to disk and refactor raw JSON logging

scharissis:
Streams the logs straight to disk so we don't use up lots of RAM.
Also implements a signal handler, so we can _try_ to exit gracefully if, say, CircleCI kills our host.

### DIFF
--- a/op-acceptor/cmd/main.go
+++ b/op-acceptor/cmd/main.go
@@ -5,6 +5,8 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"os/signal"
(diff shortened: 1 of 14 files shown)


serpixel (approved):
My only concern is that we don't have a rotatory set of logs. This can cause eventually the volume to become full, which will break op-acceptor.

scharissis:
Thanks for the comment Joe.
I won't implement it as part of this PR/change as it should probably live outside the service.

PR #534, opened 2025-12-23
feat(proxyd): add healthcheck and drain functionality from Uniswap fork

charlesma4:
**Description**

- Add configurable HTTP probe workers for backend health checking
- Backends with failing probes are excluded from load balancing
- Add graceful drain support: /healthz returns 503 when draining
- Probe config: probe_url, probe_*_threshold, probe_period_seconds

Original PR description:
```
Restarting op-stack components can generate unnecessary error noise because traffic may still be routed to containers that are about to be stopped. To ensure a cleaner shutdown process for proxyd and backend services, we’ve introduced an optional probe URL for each backend. When the probe fails, the backend is immediately removed from load balancing.

Additionally, a shutdown drain mechanism has been implemented. Upon receiving a SIGTERM, the probe will fail immediately, and the shutdown process will be delayed by 7 seconds. This gives the load balancer enough time to detect the probe failure and remove the backend from rotation before the container shuts down.

We assume the load balancer is configured with a probe interval of no more than 5 seconds, a failure threshold of 1, and a timeout of 1 second.
```

**Tests**

Added testing for common healthcheck results based on server responses - success, failure, and redirect

**Additional context**

From original PR:
```
When launching proxyd via Docker Compose, the default stop_grace_period is 10 seconds. This should be increased to at least 15 seconds—or ideally 30 seconds—to align with the default values used by Kubernetes and ECS. Otherwise, the container risks being forcefully terminated with a SIGKILL before the drain logic has a chance to complete.
```

**Metadata**

- Fixes #[Link to Issue]


proxyd/healthcheck.go
```diff
@@ -0,0 +1,177 @@
...
+		return nil, err
+	}
+
+	req.Header = http.Header{
+		"User-Agent": {"proxyd-probe"},
+		"Accept":     {"*/*"},
+	}
+
+	transport := &http.Transport{
+		TLSClientConfig:     &tls.Config{InsecureSkipVerify: true},
```
jelias2: Insecure TLS defaults to true is hardcoded which may not be desired. Can this be made a configurable value, which should default to `InsecureSkipVerify: false`

proxyd/server.go
```diff
@@ -258,6 +261,13 @@ func (s *Server) WSListenAndServe(host string, port int) error {
 	return s.wsServer.ListenAndServe()
 }
 
+func (s *Server) Drain() {
+	s.srvMu.Lock()
+	s.isDraining = true
+	time.Sleep(gracefulShutdownDuration)
+	defer s.srvMu.Unlock()
+}
```
jelias2: This appears to hold the lock for the entirety of gracefulShutdownDuration ~10s. Would it be better to update the variable and then sleep? Its possible that other threads may be blocked on s.srvMu. 

Also it might make sense consider looking at atomic.bool in this case for thread safety on a simple boolean

```suggestion
	s.srvMu.Lock()
	s.isDraining = true
	s.srvMu.Unlock()
	time.Sleep(gracefulShutdownDuration)
\
}
```

proxyd/server.go
```diff
@@ -273,6 +283,10 @@ func (s *Server) Shutdown() {
 }
 
 func (s *Server) HandleHealthz(w http.ResponseWriter, r *http.Request) {
+	if s.isDraining {
```
jelias2: `s.isDraining` is reading from the isDraining without using the mutex. Could this cause a data-race?

proxyd/proxyd.go
```diff
@@ -236,7 +240,41 @@ func Start(config *Config) (*Server, func(), error) {
...
+			if probeDelay > maxProbeDelay {
+				maxProbeDelay = probeDelay
+			}
+			// Create and launch probe worker
+			back.ProbeWorker, err = NewProbeWorker(back.probeURL, *back.probeSpec, func(healthy bool, msg string) {
+				if healthy {
+					if !back.HealthyProbe {
+						log.Info("backend is now healthy", "name", back.Name)
+					}
+					back.HealthyProbe = true
```
jelias2: The field backemd.HealthyProbe is read/written to with mutlple threads in the ProbeWorker and in backend.go without syncrhonization which could cause races, would suggest using a mutex, atomic.bool or some syncrhonization logic

proxyd/backend_probe.go
jelias2, line 1: Can healthcheck.go be renamed to ProbeSpec.go or BackendProbe.go  to more accurately reflect the logic of the file.

proxyd/backend_probe.go
```diff
@@ -0,0 +1,177 @@
...
+	if w.lastResult == result {
+		w.resultRun++
+	} else {
+		w.lastResult = result
+		w.resultRun = 1
+	}
+
+	if (!result && w.resultRun < int(w.spec.FailureThreshold)) ||
+		(result && w.resultRun < int(w.spec.SuccessThreshold)) {
+		// Success or failure is below threshold - leave the probe state unchanged.
```
jelias2, line 204: Is there empty logic here for the Failure or Success Threshold conditionals? 
Currently looks like the the latest result will always be returned

proxyd/backend_probe.go
```diff
@@ -0,0 +1,177 @@
...
+	default: // Non-blocking.
+	}
+}
+
+// Start starts the probe worker.
+func (w *ProbeWorker) Start() {
+	go w.run()
+}
+
+func (w *ProbeWorker) doProbe() {
```
jelias2, line 182: Could prometheus metrics be added to the ProbeWorker?

Something like: 
*Gauge -`backendProbeHealthy`, track 1 for healty, 0 for unhealthy 
* Counter - `backendProbeChecksTotal` - tracks number of success and failures ex. backend_probe_checks_total{backend_name="x", result="suceess"} - 1, backend_probe_checks_total{backend_name="x", result="suceess"} - 2
* Histogram -  `backendProbeDuration` to track latency of probe cehcs

proxyd/healthcheck_test.go
jelias2, line 1: Would like to see a couple additional test cases.
Test cases to validate
* Threshold behavior of ProbeWorker.doProbe
* Implement a couple more status codes ex. 503, various 40Xs,
* Testcase for DNS lookup failures
* Testcase for connection refused

proxyd/backend_probe.go
jelias2, line 1: Can some documentation be added to this file so for future uses its apparent what it does

jelias2 (changes requested):
Thanks for the contribution, look forward to landing this!

proxyd/server.go
charlesma4: fixed with atomic bool

proxyd/backend_probe.go
charlesma4: should return early here, thanks

proxyd/config.go
```diff
@@ -114,6 +114,13 @@ type BackendConfig struct {
 	StripTrailingXFF bool              `toml:"strip_trailing_xff"`
 	Headers          map[string]string `toml:"headers"`
 
+	ProbeFailureThreshold   int    `toml:"probe_failure_threshold"`
+	ProbeSuccessThreshold   int    `toml:"probe_success_threshold"`
+	ProbePeriodSeconds      int    `toml:"probe_period_seconds"`
+	ProbeTimeoutSeconds     int    `toml:"probe_timeout_seconds"`
+	ProbeURL                string `toml:"probe_url"`
+	ProbeInsecureSkipVerify bool   `toml:"probe_insecure_skip_verify"`
```
jelias2: Implementation for ProbeInsecureSkipVerify seems to be missing

proxyd/backend.go
```diff
@@ -895,6 +932,14 @@ func (b *Backend) IsHealthy() bool {
...
 }
 
+func (b *Backend) IsProbeHealthy() bool {
+	return b.healthyProbe.Load()
+}
+
+func (b *Backend) SetProbeHealth(healthy bool) {
+	b.healthyProbe.Store(healthy)
+}
+
```
jelias2, line 954:
```
func (b *Backend) IsProbeHealthy() bool {
	// If no probe is configured, return true (can't determine probe health without a probe)
	if b.probeSpec == nil {
		return true
	}
	return b.healthyProbe.Load()
}

func (b *Backend) SetProbeHealth(healthy bool) {
	// Only set probe health if a probe is configured
	if b.probeSpec != nil {
		b.healthyProbe.Store(healthy)
	}
}
```

jelias2: I think this should help with the failing tests

proxyd/backend.go
```diff
@@ -884,6 +918,9 @@ func (b *Backend) doForward(ctx context.Context, rpcReqs []*RPCReq, isBatch bool
 
 // IsHealthy checks if the backend is able to serve traffic, based on dynamic parameters
 func (b *Backend) IsHealthy() bool {
+	if !b.healthyProbe.Load() {
+		return false
+	}
```
jelias2:
```suggestion
	if !b.IsProbeHealthy() {
		return false
	}
```

jelias2: Reuse IsProbeHealthy to avoid scenarios where the probeSpec is nil

proxyd/backend.go
charlesma4: ah thought the tests were due to multicall being flaky, my bad

proxyd/config.go
charlesma4: forgot to save this file 🤦 , thanks for catching

jelias2 (approved):

PR #538, opened 2026-01-13
feat(proxyd): rate limit exempt API keys

charlesma4:
**Description**

Introduce `API_KEYS` env var to allow clients that specify an allow-listed API key to bypass rate limits. Can be extended to future compliance-related gates as well.

**Tests**

Integration test to bypass rate limits when matching API key is included in header by client.

**Additional context**

Needed for partners like QN/Alchemy to send transactions without hitting rate limits.

### DIFF
--- a/proxyd/proxyd.go
+++ b/proxyd/proxyd.go
@@ -418,9 +418,9 @@ func Start(config *Config) (*Server, func(), error) {
 	}
 
 	// Allow API Keys that bypass rate-limiting.
-	apiKeys := []string{}
+	var apiKeys []string
 	if keys, err := ReadFromEnvOrConfig("$API_KEYS"); err == nil {
-		apiKeys = append(apiKeys, strings.Split(keys, ",")...)
+		apiKeys = parseCommaSeparatedList(keys)
 	}
 
 	srv, err := NewServer(
@@ -607,6 +607,17 @@ func validateReceiptsTarget(val string) (string, error) {
 
 func secondsToDuration(seconds int) time.Duration {
 	return time.Duration(seconds) * time.Second
+}
+
+func parseCommaSeparatedList(input string) []string {
+	var result []string
+	for _, item := range strings.Split(input, ",") {
+		item = strings.TrimSpace(item)
+		if item != "" {
+			result = append(result, item)
+		}
+	}
+	return result
 }
 
 func millisecondsToDuration(ms int) time.Duration {
--- a/proxyd/proxyd.go
+++ b/proxyd/proxyd.go
@@ -9,6 +9,7 @@ import (
 	"math"
 	"net/http"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/ethereum/go-ethereum/log"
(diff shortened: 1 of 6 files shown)


proxyd/proxyd.go
```diff
@@ -416,6 +417,18 @@ func Start(config *Config) (*Server, func(), error) {
...
 	}
 
+	// Allow API Keys that bypass rate-limiting.
+	apiKeys := []string{}
+	if qnAPIKey, err := ReadFromEnvOrConfig("$QUICKNODE_API_KEY"); err == nil {
+		apiKeys = append(apiKeys, qnAPIKey)
+	}
+	if alchAPIKey, err := ReadFromEnvOrConfig("$ALCHEMY_API_KEY"); err == nil {
+		apiKeys = append(apiKeys, alchAPIKey)
+	}
```
yashvardhan-kukreja: These env vars opinionate proxyd to Quicknode, Alchemy.

Can we get rid of them, and only rely on the `$API_KEYS` env var?

proxyd/server.go
```diff
@@ -578,7 +592,7 @@ func (s *Server) handleBatchRPC(ctx context.Context, reqs []json.RawMessage, isL
 		// Apply a sender-based rate limit if it is enabled. Note that sender-based rate
 		// limits apply regardless of origin or user-agent. As such, they don't use the
 		// isLimited method.
-		if parsedReq.Method == "eth_sendRawTransaction" || parsedReq.Method == "eth_sendRawTransactionConditional" {
+		if (parsedReq.Method == "eth_sendRawTransaction" || parsedReq.Method == "eth_sendRawTransactionConditional") && !bypassLimit {
 			tx, err := s.convertSendReqToSendTx(ctx, parsedReq)
```
yashvardhan-kukreja, line 594: This check would be skipped if `bypassLimit` is true which shouldn't be the case.

Can we have the `if !bypassLimit` before or within `s.rateLimitSender` ?

proxyd/server.go
```diff
@@ -285,6 +288,13 @@ func (s *Server) HandleRPC(w http.ResponseWriter, r *http.Request) {
 	ctx, cancel = context.WithTimeout(ctx, s.timeout)
 	defer cancel()
 
+	// User can provide an API key either via the "X-Api-Key" header or by appending it to the URL path.
+	apiKey := r.Header.Get("X-Api-Key")
+	if apiKey == "" {
+		apiKey = strings.Trim(r.URL.Path, "/")
+	}
```
yashvardhan-kukreja: Any strong reason to allow including the api-key in the url?

Usually, it's more of a bad practice imo given the higher probability of leaking api keys via a URL (through logs, referrer headers, etc.)

Wouldn't ideally want proxyd to suggest such kinda practice.
- I can maybe guess some scenarios where URL baked api-keys might be convenient but I don't think if we really need to address those cases yet.

So, can we just stick to expecting api-keys only in the header?

proxyd/proxyd.go
```diff
@@ -416,6 +417,12 @@ func Start(config *Config) (*Server, func(), error) {
 		return nil, nil, fmt.Errorf("invalid interop validating strategy: %s", config.InteropValidationConfig.Strategy)
 	}
 
+	// Allow API Keys that bypass rate-limiting.
+	apiKeys := []string{}
+	if keys, err := ReadFromEnvOrConfig("$API_KEYS"); err == nil {
+		apiKeys = append(apiKeys, strings.Split(keys, ",")...)
+	}
+
```
jelias2, line 425: Would recommend filtering for empty strings, and also trimming whitespace from the split keys. 

If a user accidentally added an extra comma or trailing `key1,,key2,` this would allow easily bypassing the ratelimit with an empty string in the header

jelias2: Adding a test case for this would be great

yashvardhan-kukreja (approved):

PR #539, opened 2026-01-20
feat(proxyd): tx validation middleware

charlesma4:
**Description**

Transaction validation middleware with batching, which allows you to specify a custom URL and interface (if needed) to pass transactions to for additional checks. Either allows the group of transactions through, or blocks the entire request (for bundles).

Middleware endpoint has static request and response interfaces, which accept a map of txHash -> txObject (with 'from')

Defaults to fail open in case the middleware returns an error code, which is interpreted as server error.

Supports `eth_sendRawTransaction | eth_sendRawTransactionConditional | eth_sendBundle` currently.

**Tests**

Full unit testing for tx validation middleware file, including all utility functions.

### DIFF
--- a/proxyd/backend.go
+++ b/proxyd/backend.go
@@ -135,7 +135,7 @@ var (
 	ErrTransactionRejected = &RPCErr{
 		Code:          JSONRPCErrorInternal - 100,
 		Message:       "transaction rejected",
-		HTTPErrorCode: 500,
+		HTTPErrorCode: 400,
 	}
 
 	ErrBackendUnexpectedJSONRPC = errors.New("backend returned an unexpected JSON-RPC response")
--- a/proxyd/backend.go
+++ b/proxyd/backend.go
@@ -132,6 +132,12 @@ var (
 		HTTPErrorCode: 429,
 	}
 
+	ErrTransactionRejected = &RPCErr{
+		Code:          JSONRPCErrorInternal - 100,
+		Message:       "transaction rejected",
+		HTTPErrorCode: 500,
+	}
+
 	ErrBackendUnexpectedJSONRPC = errors.New("backend returned an unexpected JSON-RPC response")
 
 	ErrConsensusGetReceiptsCantBeBatched = errors.New("consensus_getReceipts cannot be batched")
--- a/proxyd/example.config.toml
+++ b/proxyd/example.config.toml
@@ -134,17 +134,23 @@ eth_blockNumber = "alchemy"
 
 # Transaction validation middleware configuration.
 # This middleware can validate transactions before they are forwarded to backends.
+# Note: API keys only bypass rate limits, not transaction validation.
 [tx_validation_middleware]
 # Enable or disable the validation middleware, default false
 enabled = false
 # The URL of the validation middleware service endpoint
 # endpoint = "https://your-validation-service.example.com/validate"
-# List of RPC methods to apply validation to, defaults to ["eth_sendRawTransaction", "eth_sendBundle"]
-# methods = ["eth_sendRawTransaction", "eth_sendBundle"]
+# List of RPC methods to apply validation to.
+# Defaults to ["eth_sendRawTransaction", "eth_sendRawTransactionConditional", "eth_sendBundle"]
+# methods = ["eth_sendRawTransaction", "eth_sendRawTransactionConditional", "eth_sendBundle"]
 # Timeout in seconds for validation HTTP requests, default 5
 # timeout_seconds = 5
+# Whether to allow transactions through if the validation service returns an error.
+# Defaults to true (fail-open) for safety - service disruptions won't block all transactions.
+# Set to false to reject transactions when the validation service is unavailable.
+# fail_open = true
 # Field mappings to transform the transaction object before sending to the middleware.
-# If empty (default), the full transaction object is sent.
+# If empty (default), the full transaction object is sent in go-ethereum's native JSON format.
 # Example: To send only the sender address in the format {"address": "<from_address>"}:
 # [[tx_validation_middleware.field_mappings]]
 # source_field = "from"
--- a/proxyd/example.config.toml
+++ b/proxyd/example.config.toml
@@ -131,3 +131,22 @@ secret = "test"
 eth_call = "main"
 eth_chainId = "main"
 eth_blockNumber = "alchemy"
+
+# Transaction validation middleware configuration.
+# This middleware can validate transactions before they are forwarded to backends.
+[tx_validation_middleware]
+# Enable or disable the validation middleware, default false
+enabled = false
+# The URL of the validation middleware service endpoint
+# endpoint = "https://your-validation-service.example.com/validate"
+# List of RPC methods to apply validation to, defaults to ["eth_sendRawTransaction", "eth_sendBundle"]
+# methods = ["eth_sendRawTransaction", "eth_sendBundle"]
+# Timeout in seconds for validation HTTP requests, default 5
+# timeout_seconds = 5
+# Field mappings to transform the transaction object before sending to the middleware.
+# If empty (default), the full transaction object is sent.
+# Example: To send only the sender address in the format {"address": "<from_address>"}:
+# [[tx_validation_middleware.field_mappings]]
+# source_field = "from"
+# target_field = "address"
+# Supported source fields: from, to, value, data, nonce, gas, gasPrice, chainId, hash, type, maxFeePerGas, maxPriorityFeePerGas
--- a/proxyd/example.config.toml
+++ b/proxyd/example.config.toml
@@ -134,7 +134,6 @@ eth_blockNumber = "alchemy"
 
 # Transaction validation middleware configuration.
 # This middleware can validate transactions before they are forwarded to backends.
-# Note: API keys only bypass rate limits, not transaction validation.
 [tx_validation_middleware]
 # Enable or disable the validation middleware, default false
 enabled = true
--- a/proxyd/example.config.toml
+++ b/proxyd/example.config.toml
@@ -137,22 +137,15 @@ eth_blockNumber = "alchemy"
 # Note: API keys only bypass rate limits, not transaction validation.
 [tx_validation_middleware]
 # Enable or disable the validation middleware, default false
-enabled = false
+enabled = true
 # The URL of the validation middleware service endpoint
-# endpoint = "https://your-validation-service.example.com/validate"
+endpoint = "https://your-validation-service.example.com/validate"
 # List of RPC methods to apply validation to.
 # Defaults to ["eth_sendRawTransaction", "eth_sendRawTransactionConditional", "eth_sendBundle"]
-# methods = ["eth_sendRawTransaction", "eth_sendRawTransactionConditional", "eth_sendBundle"]
+methods = ["eth_sendRawTransaction", "eth_sendRawTransactionConditional", "eth_sendBundle"]
 # Timeout in seconds for validation HTTP requests, default 5
-# timeout_seconds = 5
+timeout_seconds = 5
 # Whether to allow transactions through if the validation service returns an error.
 # Defaults to true (fail-open) for safety - service disruptions won't block all transactions.
 # Set to false to reject transactions when the validation service is unavailable.
-# fail_open = true
-# Field mappings to transform the transaction object before sending to the middleware.
-# If empty (default), the full transaction object is sent in go-ethereum's native JSON format.
-# Example: To send only the sender address in the format {"address": "<from_address>"}:
-# [[tx_validation_middleware.field_mappings]]
-# source_field = "from"
-# target_field = "address"
-# Supported source fields: from, to, value, data, nonce, gas, gasPrice, chainId, hash, type, maxFeePerGas, maxPriorityFeePerGas
+fail_open = true
--- a/proxyd/proxyd.go
+++ b/proxyd/proxyd.go
@@ -253,8 +253,11 @@ func Start(config *Config) (*Server, func(), error) {
 		log.Info("tx validation middleware enabled",
 			"endpoint", config.TxValidationMiddlewareConfig.Endpoint,
 			"methods", config.TxValidationMiddlewareConfig.Methods,
-			"field_mappings_count", len(config.TxValidationMiddlewareConfig.FieldMappings),
 		)
+		if len(config.TxValidationMiddlewareConfig.Methods) == 0 {
+			log.Warn("tx_validation_middleware enabled without explicit methods config, using defaults",
+				"default_methods", DefaultTxValidationMethods)
+		}
 	}
 
 	if config.InteropValidationConfig.LoadBalancingUnhealthinessTimeout == 0 && config.InteropValidationConfig.Strategy == HealthAwareLoadBalancingStrategy {
--- a/proxyd/proxyd.go
+++ b/proxyd/proxyd.go
@@ -245,6 +245,18 @@ func Start(config *Config) (*Server, func(), error) {
 		config.InteropValidationConfig.Strategy = defaultInteropValidationStrategy
 	}
 
+	if config.TxValidationMiddlewareConfig.Enabled && config.TxValidationMiddlewareConfig.Endpoint == "" {
+		return nil, nil, errors.New("tx_validation_middleware is enabled but no endpoint is configured")
+	}
+
+	if config.TxValidationMiddlewareConfig.Enabled {
+		log.Info("tx validation middleware enabled",
+			"endpoint", config.TxValidationMiddlewareConfig.Endpoint,
+			"methods", config.TxValidationMiddlewareConfig.Methods,
+			"field_mappings_count", len(config.TxValidationMiddlewareConfig.FieldMappings),
+		)
+	}
+
 	if config.InteropValidationConfig.LoadBalancingUnhealthinessTimeout == 0 && config.InteropValidationConfig.Strategy == HealthAwareLoadBalancingStrategy {
 		log.Warn("no interop validation load balancing unhealthiness timeout provided for health aware strategy, using default timeout", "timeout", defaultInteropLoadBalancingUnhealthinessTimeout)
 		config.InteropValidationConfig.LoadBalancingUnhealthinessTimeout = defaultInteropLoadBalancingUnhealthinessTimeout
@@ -446,6 +458,7 @@ func Start(config *Config) (*Server, func(), error) {
 		interopStrategy,
 		config.Server.EnableTxHashLogging,
 		apiKeys,
+		config.TxValidationMiddlewareConfig,
 	)
 	if err != nil {
 		return nil, nil, fmt.Errorf("error creating server: %w", err)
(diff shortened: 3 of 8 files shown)


proxyd/backend.go
```diff
@@ -132,6 +132,12 @@ var (
 		HTTPErrorCode: 429,
 	}
 
+	ErrTransactionRejected = &RPCErr{
+		Code:          JSONRPCErrorInternal - 100,
+		Message:       "transaction rejected",
+		HTTPErrorCode: 500,
```
yashvardhan-kukreja: is the 500 for obfuscating the error to the potentially malicious client?

If not, I'd suggest a bad request http code to be used.

proxyd/example.config.toml
```diff
@@ -131,3 +131,22 @@ secret = "test"
...
+
+# Transaction validation middleware configuration.
+# This middleware can validate transactions before they are forwarded to backends.
+[tx_validation_middleware]
+# Enable or disable the validation middleware, default false
+enabled = false
+# The URL of the validation middleware service endpoint
+# endpoint = "https://your-validation-service.example.com/validate"
+# List of RPC methods to apply validation to, defaults to ["eth_sendRawTransaction", "eth_sendBundle"]
+# methods = ["eth_sendRawTransaction", "eth_sendBundle"]
```
yashvardhan-kukreja:
```suggestion
# List of RPC methods to apply validation to, defaults to ["eth_sendRawTransaction", "eth_sendRawTransactionConditional",  "eth_sendBundle"]
# methods = ["eth_sendRawTransaction", "eth_sendRawTransactionConditional", "eth_sendBundle"]
```

proxyd/proxyd.go
```diff
@@ -245,6 +245,18 @@ func Start(config *Config) (*Server, func(), error) {
...
+		return nil, nil, errors.New("tx_validation_middleware is enabled but no endpoint is configured")
+	}
+
+	if config.TxValidationMiddlewareConfig.Enabled {
+		log.Info("tx validation middleware enabled",
+			"endpoint", config.TxValidationMiddlewareConfig.Endpoint,
+			"methods", config.TxValidationMiddlewareConfig.Methods,
+			"field_mappings_count", len(config.TxValidationMiddlewareConfig.FieldMappings),
+		)
+	}
```
yashvardhan-kukreja, line 261: Would be better to log.Warn about falling back to the default methods given if len(methods) == 0

proxyd/tx_validation_middleware.go
```diff
@@ -0,0 +1,368 @@
...
+		set[m] = struct{}{}
+	}
+	return set
+}
+
+// defaultTxValidationMethods returns the default set of methods to validate.
+func defaultTxValidationMethods() TxValidationMethodSet {
+	return TxValidationMethodSet{
+		"eth_sendRawTransaction": {},
+		"eth_sendBundle":         {},
```
yashvardhan-kukreja:
```suggestion
		"eth_sendBundle":         {},
        "eth_sendRawTransactionConditional": {},
```

based on one of the previous suggestions.

proxyd/tx_validation_middleware.go
```diff
@@ -0,0 +1,368 @@
...
+}
+
+// txValidation is the default validation function using the shared client.
+func txValidation(ctx context.Context, endpoint string, payload []byte) (bool, error) {
+	return getDefaultTxValidationClient().Validate(ctx, endpoint, payload)
+}
+
+// txValidationResponse represents the response from the validation middleware.
+type txValidationResponse struct {
+	Block        bool   `json:"block"`
```
yashvardhan-kukreja: nit: "block" can have multiple meanings in web3. I'd prefer a more distinct term like "success" or "authorized" (or "failed" / "unauthorized" with a negative connotation)

proxyd/server.go
```diff
@@ -209,19 +229,25 @@ func NewServer(
...
+		allowedChainIds:           senderRateLimitConfig.AllowedChainIds,
+		limExemptOrigins:          limExemptOrigins,
+		limExemptUserAgents:       limExemptUserAgents,
+		limExemptKeys:             limExemptKeys,
+		rateLimitHeader:           rateLimitHeader,
+		interopValidatingConfig:   interopValidatingConfig,
+		interopStrategy:           interopStrategy,
+		enableTxHashLogging:       enableTxHashLogging,
+		enableTxValidation:        txValidationConfig.Enabled,
+		txValidationFn:            txValidation,
```
yashvardhan-kukreja: Any reason to have this txValidationFn given we already are forming the txValidationClient, and for the caller, the methods signatures are equally abstracted (same inputs) for txValidationFn and txValidationClient.Validate(...)?

yashvardhan-kukreja: And if we really wanna use it, then it'd be better to skip the "func txValidation", and just set `txValidationFn: txValidationClient.Validate`

proxyd/server.go
```diff
@@ -610,7 +636,15 @@ func (s *Server) handleBatchRPC(ctx context.Context, reqs []json.RawMessage, isL
 				responses[i] = NewRPCErrorRes(parsedReq.ID, err)
 				continue
 			}
+		}
 
+		// Apply transaction validation middleware if enabled and method is configured
+		if s.enableTxValidation && !bypassLimit && s.txValidationMethods.Contains(parsedReq.Method) {
```
yashvardhan-kukreja: The semantics of bypassLimit pertain to rate-limit exemption only. I don't think that flag should be used to skip transaction validation. So, can we remove `!bypassLimit`?

proxyd/tx_validation_middleware.go
```diff
@@ -0,0 +1,368 @@
...
+	fieldMappings []TxFieldMapping,
+	validationFn TxValidationFunc,
+) error {
+	if len(txs) > maxBundleTransactions {
+		log.Warn("bundle contains too many transactions",
+			"req_id", GetReqID(ctx),
+			"tx_count", len(txs),
+			"max_allowed", maxBundleTransactions)
+		return ErrInvalidParams(fmt.Sprintf("bundle contains %d transactions, maximum allowed is %d", len(txs), maxBundleTransactions))
+	}
```
yashvardhan-kukreja, line 172: I think this kind of logic (len(txs) <= maxBundleTransactions) either should be configurable or even better be programmed on the validation middleware. (especially considering it can be workedaround a bit by playing with max-request-body-size param of proxyd).

proxyd/tx_validation_middleware.go
```diff
@@ -0,0 +1,368 @@
...
+		return buildFullTxPayload(tx, from)
+	}
+	return buildMappedPayload(tx, from, fieldMappings)
+}
+
+// creates a payload containing the full transaction details, including 'from' address
+func buildFullTxPayload(tx *types.Transaction, from common.Address) ([]byte, error) {
+	payload := map[string]interface{}{
+		"tx":   tx,
+		"from": from.Hex(),
```
yashvardhan-kukreja: This is redundant as the tx already holds the from address. Also opinionates the validation towards sender-address-based which can be a wrong assumption. (some validators might incline more towards receiver address or/and calldata addresses).

I think the fulltxPayload should be completely sent as is in the form of `types.Transaction`, keeps the interface more predictable for the external validator

proxyd/tx_validation_middleware.go
```diff
@@ -0,0 +1,368 @@
...
+	}
+	return json.Marshal(payload)
+}
+
+// creates a payload with only the mapped fields.
+func buildMappedPayload(tx *types.Transaction, from common.Address, mappings []TxFieldMapping) ([]byte, error) {
```
yashvardhan-kukreja: imo, we should only send the full transaction as is to the middleware. The mapped payload is complicating the interface for both the proxyd owner (makes the validation config more imperative than declarative) as well as the validation-middleware's owner. 
- The client (middleware) should simply succumb to the interface offered by the Server (proxyd) while the server continues to keep that interface flexibly straightforward (types.Transaction).

The only place where I can see mapped payload being helpful is to reduce the transaction fields egress-ing but imo that sounds like a premature optimization at this moment.
- And there are better ways to skip egressing fields while continuing to send types.Transaction (keeping the interface same) even if we wanna go in that direction but I'd prefer not to at this moment

proxyd/tx_validation_middleware.go
```diff
@@ -0,0 +1,368 @@
...
+		if value != nil {
+			payload[mapping.TargetField] = value
+		}
+	}
+
+	return json.Marshal(payload)
+}
+
+// extractTxField extracts a field value from the transaction.
+func extractTxField(tx *types.Transaction, from common.Address, field string) interface{} {
```
yashvardhan-kukreja: This functions requires leaking the abstraction of types.Transaction. (another reason inclining against mapped payload). If we continue to use mapped payload (my previous comment), it would be less leaky to just json marshal->unmarshal the txn into a `res := map[string]interface{}` and accordingly, just parse out `res[field]` (with a little bit of logic like GasFeeCap and GasTipCap)

proxyd/example.config.toml
yashvardhan-kukreja: I think it would be more visually appealing to have `enabled = true` and all the functional fields uncommented (leaving only the descriptions commented).

wdyt?

proxyd/tx_validation_middleware.go
```diff
@@ -0,0 +1,368 @@
...
+			"req_id", GetReqID(ctx),
+			"from", from.Hex(),
+			"error", validationErr,
+			"chain_id", tx.ChainId(),
+			"nonce", tx.Nonce(),
+			"value", tx.Value(),
+			"tx_hash", tx.Hash().Hex(),
+		)
+		return nil
+	}
```
yashvardhan-kukreja: The fail open behaviour can be undesired for a bunch of users seeking some critical validation. But I like this direction.

Let's have a fail_open (or `allow_on_error`) arg determining whether to fail-open / fail-close. The default of it can be to fail open.

proxyd/server.go
charlesma4: yes can remove, originally left in for perf purposes as we assumed trusted clients to validate their own txs

proxyd/proxyd.go
charlesma4: like an additional log.warn? down to add that

proxyd/tx_validation_middleware.go
charlesma4: can make it configurable

transaction list length is a request-level field though, so I think it makes sense to have proxyd validate this, not the tx validation middleware (which validates tx-level params)

proxyd/server.go
yashvardhan-kukreja: Yep makes sense but `bypassLimit` is probably not worth reusing here. Ideally I would expect the middleware to ignore the validation wherever necessary (the extra IO shouldn't hurt the trusted addresses :P )

proxyd/tx_validation_middleware.go
yashvardhan-kukreja: Would you mind creating a separate PR for it (feel free to check it out against `main` or your current branch, your choice). Would lighten this PR.

proxyd/tx_validation_middleware.go
charlesma4: I tested it, and it seems the `from` address is not included during json marshaling. We could ask the validation service to recalculate the `from` address using `ecrecover(hash, v, r, s)` but I feel it's reasonable to include `from` in the payload for convenience

```
{
  "tx": {
    "type": "0x2",
    "chainId": "0x1",
    "nonce": "0x1",
    "to": "0x0987654321098765432109876543210987654321",
    "gas": "0x5208",
    "maxPriorityFeePerGas": "0x3b9aca00",
    "maxFeePerGas": "0x3b9aca00",
    "value": "0xde0b6b3a7640000",
    "input": "0x0102",
    "accessList": [],
    "v": "0x1",
    "r": "0x357eb2d6f5fc421709cf216d3f7f2bcbc5d103465c5f1f3a290eb21f14b0d974",
    "s": "0x74102e7167f1f2e1e3cdf76cc6e3619dd49b2227ee309a423566b8443be16dba",
    "yParity": "0x1",
    "hash": "0x714e7a31a2f382888c2169a38639848e7c1e8b5dc5f7f1d3266f8b6ff02ecaed"
  }
}
```

proxyd/tx_validation_middleware.go
charlesma4: switching to `unauthorized`

proxyd/tx_validation_middleware.go
charlesma4: decided to put `from` address inside the tx object for the request payload

proxyd/tx_validation_middleware.go
```diff
@@ -0,0 +1,297 @@
...
+// txValidationResponse represents the response from the validation middleware.
+type txValidationResponse struct {
+	Unauthorized map[string]bool `json:"unauthorized"`
+	ErrorCode    string          `json:"errorCode"`
+	ErrorMessage string          `json:"errorMessage"`
+}
+
+// txValidationRequest is the request payload sent to the validation service.
+type txValidationRequest struct {
+	Txs map[string]map[string]interface{} `json:"txs"`
```
yashvardhan-kukreja: <del>Can we use `map[string]types.Transaction`, especially considering that the intention is to send a directly parseable vanilla transaction object. Also, the fact the `types` is coming from the official go-ethereum library makes this more reliably concrete and maintained.</del>

<del>wdyt?</del>

<del>PS: if there are situations (certain methods for example) where one would not want the proxyd to send the fully populated `types.Transaction` object (say, with calldata), then there's anyway always a way for them to contribute to proxyd allowing to nullify those fields programmatically.</del>

proxyd/tx_validation_middleware.go
```diff
@@ -0,0 +1,297 @@
...
+		return err
+	}
+
+	payload, err := buildValidationPayload(txsWithSenders)
+	if err != nil {
+		log.Error("error building validation payload", "err", err, "req_id", GetReqID(ctx))
+		return ErrInternal
+	}
+
+	unauthorized, validationErr := validationFn(ctx, endpoint, payload)
```
yashvardhan-kukreja:
```suggestion
	validationResults, validationErr := validationFn(ctx, endpoint, payload)
```

nit

proxyd/tx_validation_middleware.go
```diff
@@ -0,0 +1,297 @@
...
+		)
+		return ErrInternal
+	}
+
+	for txHash, isUnauthorized := range unauthorized {
+		if isUnauthorized {
+			txData := txsWithSenders[txHash]
+			log.Info("transaction rejected by validation middleware",
+				"req_id", GetReqID(ctx),
+				"from", txData["from"],
```
yashvardhan-kukreja, line 208: <del>I think we shouldn't necessarily log the "from" field especially given that we're logging the transaction hash. Moreover, a transaction can be rejected by reasons/fields much beyond the "from" address making this log potentially confusing.</del>

------
nvm, just played a bit and came to know the UnmarshalJSON doesn't unmarshal the "from" address.

So this is good then.

proxyd/tx_validation_middleware.go
```diff
@@ -0,0 +1,297 @@
...
+			)
+			return ErrTransactionRejected
+		}
+	}
+	return nil
+}
+
+// buildTxsWithSenders builds a map of tx hashes to flattened tx objects with "from" field added.
+func buildTxsWithSenders(ctx context.Context, txs []*types.Transaction) (map[string]map[string]interface{}, error) {
+	result := make(map[string]map[string]interface{}, len(txs))
```
yashvardhan-kukreja: let's try to simply form a map of `txHash` -> `txObject`, instead of `txHash` -> `txObject + from`

Considering the "from" to be easily derivable from txObject, we shouldn't need to or opinionate towards sending the "from" field explicitly.

As a reader, this ends up insinuating the validation to be from-address inclined, and convolutes the request interface for the middleware further, as compared to merely sending a raw `txObject`

proxyd/tx_validation_middleware.go
yashvardhan-kukreja: Can we create a type alias here

```
type TransactionWithFrom map[string]interface{}
```

And if you can add a comment or two about why this alias is needed despite the existing `types.Transaction` type and its issues with not unmarshalling "from" address.

proxyd/tx_validation_middleware.go
```diff
@@ -0,0 +1,297 @@
...
+		}
+
+		var txMap map[string]interface{}
+		if err := json.Unmarshal(txJSON, &txMap); err != nil {
+			log.Debug("could not unmarshal transaction for validation", "err", err, "req_id", GetReqID(ctx))
+			return nil, ErrInternal
+		}
+
+		// Add "from" at the same level as other tx fields
+		txMap["from"] = from.Hex()
```
yashvardhan-kukreja, line 243: Just a small comment here as well (probably duplicate but worthwhile) about the need of doing so because of the absence of unmarshalling of "from" field.

proxyd/tx_validation_middleware_test.go
```diff
@@ -0,0 +1,401 @@
...
+	require.NoError(t, err)
+
+	txsWithSenders, err := buildTxsWithSenders(context.Background(), []*types.Transaction{tx})
+	require.NoError(t, err)
+
+	payload, err := buildValidationPayload(txsWithSenders)
+	require.NoError(t, err)
+
+	var result struct {
+		Txs map[string]map[string]interface{} `json:"txs"`
```
yashvardhan-kukreja:
```suggestion
		Txs map[string]TransactionWithFrom `json:"txs"`
```

proxyd/tx_validation_middleware_test.go
```diff
@@ -0,0 +1,401 @@
...
+	for i := 0; i < 3; i++ {
+		txs[i] = createSignedTestTransaction(t)
+	}
+
+	callCount := 0
+	mockValidation := func(ctx context.Context, endpoint string, payload []byte) (map[string]bool, error) {
+		callCount++
+		// Verify the payload contains all 3 txs wrapped in "txs" field
+		var request struct {
+			Txs map[string]map[string]interface{} `json:"txs"`
```
yashvardhan-kukreja:
```suggestion
			Txs map[string]TransactionWithFrom `json:"txs"`
```

yashvardhan-kukreja (approved):
lgtm

Thanks @charlesma4

proxyd/tx_validation_middleware.go
```diff
@@ -0,0 +1,306 @@
...
+func buildTxsWithSenders(ctx context.Context, txs []*types.Transaction) (map[string]TransactionWithFrom, error) {
+	result := make(map[string]TransactionWithFrom, len(txs))
+	for _, tx := range txs {
+		from, err := getSender(tx)
+		if err != nil {
+			log.Debug("could not get sender from transaction for validation", "err", err, "req_id", GetReqID(ctx))
+			return nil, ErrInvalidParams(err.Error())
+		}
+
+		// Marshal tx to JSON, then unmarshal to map to get all fields
```
yashvardhan-kukreja:
```suggestion
		// Marshal txn to JSON, then unmarshal to map to get all fields
```

yashvardhan-kukreja: noop to trigger a new commit for CI pipelines

PR #547, opened 2026-02-06
fix(proxyd): forward out-of-range block requests to backend

### LINKED ISSUE #546: proxyd: Wrongly returns HTTP and JSON error if block not found
`proxyd` seems to respond with errors if a block requested with `eth_getBlockByHash` or `eth_getBlockByNumber` cannot be found. But the [Ethereum specs-compliant](https://ethereum.org/developers/docs/apis/json-rpc/#eth_getblockbyhash) response should just be a null header and no error.

E.g. getting
```
"400 Bad Request: {\"jsonrpc\":\"2.0\",\"error\":{\"code\":-32019,\"message\":\"block is out of range\"},\"id\":33549}\n"
```
for a future block, so not only is it returning a JSON error (with code `-32019`), but also a `400` HTTP error code.

Some related code in proxyd is
https://github.com/ethereum-optimism/infra/blob/f9bf450f8185bed5d12923f8fea242487dba8372/proxyd/rewriter.go#L299-L301
and
https://github.com/ethereum-optimism/infra/blob/f9bf450f8185bed5d12923f8fea242487dba8372/proxyd/backend.go#L1797-L1798

The `ethclient` also won't [properly translate](https://github.com/ethereum-optimism/op-geth/blob/c277d6773bc3f275c4cd8f94111ea45e015132c5/ethclient/ethclient.go?plain=1#L250-L252) those errors, so it's not returning an `ethereum.NotFound` error as expected for blocks that don't exist (yet).

Edit: and more generally, proxyd catches all future block number requests and returns an HTTP 400 plus JSON error, which for many RPCs is wrong, e.g. also for `eth_getStorageAt` which returns `0x0000000000000000000000000000000000000000000000000000000000000000`, not errors.

sebastianst:
## Summary

Fixes #546.

- Removes the `ErrRewriteBlockOutOfRange` / `ErrBlockOutOfRange` errors entirely
- Requests for blocks beyond `latest` are now forwarded to the backend instead of being intercepted by the rewriter
- The backend returns the spec-compliant response (e.g. `{"result": null}` for `eth_getBlockByNumber` with a future block)

Previously, proxyd returned HTTP 400 with JSON-RPC error code `-32019` ("block is out of range") for methods like `eth_getBlockByNumber` when the requested block was beyond `latest`. Per the [Ethereum JSON-RPC spec](https://ethereum.org/developers/docs/apis/json-rpc#eth_getblockbyhash), these methods should return `null`, not an error. This also broke `ethclient` which expects `ethereum.NotFound` for missing blocks, not an RPC error.

The rewriter's job is to translate block tags (`latest`, `safe`, `finalized`) to concrete numbers and enforce max block ranges — not to decide whether a specific block number exists. That's the backend's responsibility.

## Test plan

- [x] Unit tests updated — out-of-range cases now expect `RewriteNone` (pass-through) instead of `RewriteOverrideError`
- [x] Integration tests updated — verify requests for future blocks are forwarded to backend and return null
- [x] Full test suite passes (`go test ./... -timeout 10m`)

### DIFF
--- a/proxyd/backend.go
+++ b/proxyd/backend.go
@@ -97,12 +97,6 @@ var (
 		Message:       "backend is currently not healthy to serve traffic",
 		HTTPErrorCode: 503,
 	}
-	ErrBlockOutOfRange = &RPCErr{
-		Code:          JSONRPCErrorInternal - 19,
-		Message:       "block is out of range",
-		HTTPErrorCode: 400,
-	}
-
 	ErrRequestBodyTooLarge = &RPCErr{
 		Code:          JSONRPCErrorInternal - 21,
 		Message:       "request body too large",
@@ -1794,9 +1788,7 @@ func (bg *BackendGroup) OverwriteConsensusResponses(rpcReqs []*RPCReq, overridde
 				req:   req,
 				res:   &res,
 			})
-			if errors.Is(err, ErrRewriteBlockOutOfRange) {
-				res.Error = ErrBlockOutOfRange
-			} else if errors.Is(err, ErrRewriteRangeTooLarge) {
+			if errors.Is(err, ErrRewriteRangeTooLarge) {
 				res.Error = ErrInvalidParams(
 					fmt.Sprintf("block range greater than %d max", rctx.maxBlockRange),
 				)
--- a/proxyd/integration_tests/consensus_test.go
+++ b/proxyd/integration_tests/consensus_test.go
@@ -752,19 +752,26 @@ func TestConsensus(t *testing.T) {
 		require.Equal(t, "0xe1", jsonMap["params"].([]interface{})[0])
 	})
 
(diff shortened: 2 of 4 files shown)


serpixel (approved):
Looks good to me, but @jelias2 may be better suited here.

geoknee (approved):

jelias2:
So I'm going to propose some changes to this. I think you found a bug, we definetly want proxyd to be eth rpc compliant where possible  so this is a bug, however we don't want to send requests to our backends if the the backend consensus group does not recognize the block range as valid. 

So I think the approraite thing to do would be keep the checks and do not forward to the backends, yet change the error response to be eth compliant and debug log the action

sebastianst:
works for me!

PR #552, opened 2026-02-12
feat: proxyd update consensus mode out of range requests to be api-spec compliant

### LINKED ISSUE #546: proxyd: Wrongly returns HTTP and JSON error if block not found

jelias2:
**Description**
* In consensus mode proxyd if requests were out of range for the consensus backend they would return an http 400 status code and error message. This PR updates proxyd to be json-rpc spec complaint by returning a compatible response to the client. This should ensure the client can properly function regardless of proxyd intercepting requests or not  

**Tests**

**Additional context**

**Metadata**

- Fixes #[Link to Issue]

### DIFF
--- a/proxyd/backend.go
+++ b/proxyd/backend.go
@@ -1788,7 +1788,39 @@ func (bg *BackendGroup) OverwriteConsensusResponses(rpcReqs []*RPCReq, overridde
 				req:   req,
 				res:   &res,
 			})
-			if errors.Is(err, ErrRewriteRangeTooLarge) {
+			if errors.Is(err, ErrRewriteBlockOutOfRange) {
+				// Return spec-compliant response for out-of-range blocks
+				// instead of forwarding to backend (fail-fast)
+				log.Debug(
+					"returning spec-compliant response for out-of-range block request",
+					"method", req.Method,
+					"req_id", req.ID,
+				)
+				switch req.Method {
+				case "eth_getBlockByNumber",
+					"eth_getBlockByHash",
+					"eth_getBlockReceipts",
+					"eth_getBlockTransactionCountByNumber",
+					"eth_getUncleCountByBlockNumber",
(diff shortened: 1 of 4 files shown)


geoknee (approved):

geoknee:
replaces https://github.com/ethereum-optimism/infra/pull/547

yashvardhan-kukreja (approved):

PR #561, opened 2026-03-04
fix(ci): install zstd for mise archive extraction in go-release job

scharissis:
## Summary
- The mise installer now ships `.tar.zst` archives, but the `ci-builder` Docker image used by the `go-release` job lacks `zstd`, causing the "Setup mise" step to fail with `zstd: Cannot exec: No such file or directory`
- Adds `apt-get install zstd` before the mise install command to fix the extraction

## Test plan
- [ ] Trigger a release build for op-acceptor and verify the "Setup mise" step passes

### DIFF
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -277,6 +277,8 @@ jobs:
       - run:
           name: Setup mise
           command: |
+            # Install zstd (needed to extract mise archive)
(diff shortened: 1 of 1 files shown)


scharissis:
Please merge when approved.

bitwiseguy (approved):

PR #565, opened 2026-03-06
fix(op-acceptor): skip RunAll package when all tests excluded by --exclude-gates

smartcontracts:
## Summary

- Fixes a CI hang caused by \`--exclude-gates flake-shake\` on RunAll packages where every test is excluded
- When all tests in a RunAll package are in \`SkipTests\`, \`go test\` is no longer invoked — the package is reported as skipped immediately
- Adds a regression test covering the all-tests-excluded case

## Root cause

When \`--exclude-gates flake-shake\` is applied, named tests in excluded gates are added to \`SkipTests\` on their package's RunAll validator. \`go test -skip\` is then passed to the test binary. If every test in a package is in \`SkipTests\`:

1. \`go test ./reqressyncdisabled -skip ...\` runs
2. \`TestMain\` is called → \`presets.DoMain\` → full devnet starts (EL snap sync between geth nodes begins)
3. \`m.Run()\` returns 0 immediately (all tests skipped by \`-skip\`)
4. \`defer p.Close()\` fires → \`n.server.Stop()\` in go-ethereum blocks waiting for active snap sync P2P connections to close → **indefinite hang**

## Fix

Before invoking \`go test\` for a RunAll validator with a non-empty \`SkipTests\` list, check whether every \`Test*\` function in the package source is in the skip set. If so, return \`TestStatusSkip\` immediately.

**Why not \`go test -list\`?** \`go test -list\` invokes the test binary, which calls \`TestMain\`. \`presets.DoMain\` starts the devnet inside \`TestMain\` _before_ \`m.Run()\` — so the pre-check would hang on the same teardown it was meant to prevent.

**Correct approach — source-level scan:**
1. \`go list -json <package>\` resolves \`TestGoFiles\` and \`XTestGoFiles\` (no build, no binary execution)
2. Scan each file with \`regexp\` for \`^func (Test\\w+)(\` declarations
3. If every found \`Test*\` function is in \`SkipTests\`, skip the package invocation entirely

This never runs \`TestMain\` or starts any test infrastructure.

## Affected scenario

The \`reqressyncdisabled\` package in gateless/memory-all mode with \`--exclude-gates flake-shake\`. Its single test (\`TestUnsafeChainNotStalling_DisabledReqRespSync\`) is in the flake-shake gate. With this fix, the package is skipped entirely instead of hanging.

### DIFF
--- a/op-acceptor/runner/runner.go
+++ b/op-acceptor/runner/runner.go
@@ -747,20 +747,40 @@ func (r *runner) runAllTestsInPackage(ctx context.Context, metadata types.Valida
 	return res, err
 }
 
-// allTestsWouldBeSkipped returns true if every top-level test function in the
-// package is present in metadata.SkipTests, meaning no tests would actually run.
-// It uses "go test -list ." to enumerate the package's test functions without
-// executing them, so it is cheap and does not start any test infrastructure.
+// testFuncPattern matches top-level Go test function declarations. We only
+// match Test* here because SkipTests comes from explicit YAML entries which
+// are invariably Test* functions; Benchmark* and Fuzz* are not skipped via
+// -skip in normal acceptance test runs.
+var testFuncPattern = regexp.MustCompile(`^func (Test\w+)\(`)
+
+// allTestsWouldBeSkipped returns true if every top-level Test* function in the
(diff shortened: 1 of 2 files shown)


scharissis (approved):
LGTM. Thanks!

PR #569, opened 2026-03-17
feat(proxyd): log sendRawTransaction forwarding with tx_hash and duration

bitwiseguy:
## Summary

- When `enable_tx_hash_logging` is enabled, proxyd now emits a `sendRawTransaction forwarded` INFO log after the backend responds
- Log includes `tx_hash`, `req_id`, `backend`, `backend_group`, and `duration_ms`
- Gated behind the same `enable_tx_hash_logging` config flag as the existing request-time log
- Works for both single and batched `eth_sendRawTransaction` / `eth_sendRawTransactionConditional` requests

This complements the existing `processing sendRawTransaction` log (fires at request time) by adding a response-time log. The `req_id` field correlates the two, giving proxyd's internal processing duration. This is used by `op-benchmark latency` to measure the `proxyd_forwarded` stage in flashblocks latency analysis.

### Implementation

The tx hash is captured during request parsing (where it was already extracted for rate limiting) into a request-scoped map. After `Forward()` returns, any sendRawTransaction requests in the batch get a forwarding log with timing. The map is local to `handleBatchRPC` and garbage collected per request — no accumulation across requests.

## Test plan

- [x] Existing `TestTxHashLogging` tests pass with new assertions for `sendRawTransaction forwarded` and `duration_ms`
- [x] `TestTxHashLoggingDisabled` verifies forwarding log does NOT appear when `enable_tx_hash_logging` is off

### DIFF
--- a/proxyd/server.go
+++ b/proxyd/server.go
@@ -568,6 +568,7 @@ func (s *Server) handleBatchRPC(ctx context.Context, reqs []json.RawMessage, isL
 	responses := make([]*RPCRes, len(reqs))
 	batches := make(map[batchGroup][]batchElem)
 	ids := make(map[string]int, len(reqs))
+	txHashes := make(map[int]common.Hash) // tracks tx hashes by request index for forwarding logs
 
 	for i := range reqs {
 		parsedReq, err := ParseRPCReq(reqs[i])
@@ -653,6 +654,7 @@ func (s *Server) handleBatchRPC(ctx context.Context, reqs []json.RawMessage, isL
 				responses[i] = NewRPCErrorRes(parsedReq.ID, err)
 				continue
 			}
+			txHashes[i] = tx.Hash()
 			if err := s.rateLimitSender(ctx, tx, bypassLimit); err != nil {
 				RecordRPCError(ctx, BackendProxyd, parsedReq.Method, err)
 				responses[i] = NewRPCErrorRes(parsedReq.ID, err)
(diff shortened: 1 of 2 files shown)


hdcesario-op (commented):
**[P2] Don't emit "forwarded" logs on failed upstream calls**

`proxyd/server.go` (lines 723-738): The new `sendRawTransaction forwarded` log block is emitted **before** the `err != nil` check on the `Forward()` result. `BackendGroup.Forward` can return errors for real failure cases — exhausted failover (`ErrBackendOffline`), `ErrBackendResponseTooLarge`, `ErrContextCanceled`, and others (see `ForwardRequestToBackendGroup` in `backend.go` line 1682+). When this happens, proxyd logs a successful-looking forwarding record even though the request ultimately fails.

This will skew the latency/audit signal that this PR is adding for `op-benchmark`.

**Suggested fix:** Gate the logging on `err == nil`, or move it into the success path after the error handling block.

---

No other high-confidence issues found in the two touched files. Existing tests pass (`TestTxHashLogging` and `TestTxHashLoggingDisabled`). The main remaining test gap is coverage for the failed-forwarding path, which would catch the issue above once fixed.

proxyd/server.go
```diff
@@ -713,8 +715,26 @@ func (s *Server) handleBatchRPC(ctx context.Context, reqs []json.RawMessage, isL
 			start := i * s.maxUpstreamBatchSize
 			end := int(math.Min(float64(start+s.maxUpstreamBatchSize), float64(len(cacheMisses))))
 			elems := cacheMisses[start:end]
+			forwardStart := time.Now()
 			res, sb, err := s.BackendGroups[group.backendGroup].Forward(ctx, createBatchRequest(elems), isBatch)
+			forwardDuration := time.Since(forwardStart)
 			servedBy[sb] = true
+
+			// Log forwarding for sendRawTransaction requests
```
hdcesario-op: proxyd/server.go (lines 723-738):  The new sendRawTransaction forwarded log block is emitted before the `err != nil` check on the Forward() result. BackendGroup.

Forward can return errors for real failure cases — exhausted failover (ErrBackendOffline), ErrBackendResponseTooLarge, ErrContextCanceled, and others (see ForwardRequestToBackendGroup in backend.go line 1682+). 

When this happens, proxyd logs a successful-looking forwarding record even though the request ultimately fails. This will skew the latency/audit signal that this PR is adding for op-benchmark.

Suggested fix: Gate the logging on `err == nil`, or move it into the success path after the error handling block.

No other high-confidence issues found in the two touched files. Existing tests pass (TestTxHashLogging and TestTxHashLoggingDisabled). The main remaining test gap is coverage for the failed-forwarding path, which would catch the issue above once fixed.

bitwiseguy: Great catch. Addressed in https://github.com/ethereum-optimism/infra/pull/569/commits/cac5167880cded32eadc2cfca08452d7f2bc6b09

hdcesario-op (approved):
LGTM

PR #576, opened 2026-03-22
feat(cci-stats): map flaky test markers to structured statuses

ajsutton:
## Summary

- Stop filtering out skipped tests — record all test results in the database for complete visibility
- Add `mapFlakyStatus()` helper that detects `FLAKY_FAIL:` and `FLAKY_PASS:` prefixes in skip messages, mapping them to `"failed"` and `"flaky_pass"` statuses respectively
- Regular skips pass through as `"skipped"`

This provides more complete test data and enables tracking flaky test outcomes from devtest's MarkFlaky annotations. Flaky failures show as `"failed"` (matching previous op-acceptor behavior), `"flaky_pass"` entries enable querying for tests that have stabilized, and regular skips are now recorded too.

No schema migration needed — `status` is VARCHAR and the existing index covers new values.

Companion PR: ethereum-optimism/optimism#19719 — devtest changes that emit the FLAKY_FAIL/FLAKY_PASS markers.

## Test plan

- [x] New test: `TestFlakyStatusMapping` — table-driven test covering all 6 cases (flaky fail, flaky pass, regular skip, empty skip, regular failure, success)
- [x] All existing tests pass
- [x] `go vet` clean

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/cci-stats/pkg/service/service.go
+++ b/cci-stats/pkg/service/service.go
@@ -264,7 +264,7 @@ func fetchJobs(ctx context.Context, client *circleci.Client, workflowID string) 
 // and all other statuses pass through unchanged.
 func mapFlakyStatus(result, message string) string {
 	if result == "skipped" {
-		if strings.HasPrefix(message, "FLAKY_FAIL:") {
+		if strings.HasPrefix(message, "FLAKY_FAIL") {
 			return "failed"
 		}
 		if strings.HasPrefix(message, "FLAKY_PASS") {
--- a/cci-stats/pkg/service/service.go
+++ b/cci-stats/pkg/service/service.go
@@ -259,6 +259,21 @@ func fetchJobs(ctx context.Context, client *circleci.Client, workflowID string) 
 	return jobs.Items, nil
 }
 
+// mapFlakyStatus maps CircleCI test result status based on flaky markers
+// in the skip message. Returns the mapped status and whether to keep the result.
+func mapFlakyStatus(result, message string) (status string, keep bool) {
+	if result == "skipped" {
+		if strings.HasPrefix(message, "FLAKY_FAIL:") {
+			return "failed", true
+		}
+		if strings.HasPrefix(message, "FLAKY_PASS:") {
+			return "flaky_pass", true
+		}
+		return "", false
+	}
+	return result, true
+}
+
 func isNotFoundError(err error) bool {
 	return err != nil && strings.Contains(strings.ToLower(err.Error()), "not found")
 }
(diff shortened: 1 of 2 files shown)


cci-stats/pkg/service/service.go
```diff
@@ -259,6 +259,21 @@ func fetchJobs(ctx context.Context, client *circleci.Client, workflowID string)
...
 
+// mapFlakyStatus maps CircleCI test result status based on flaky markers
+// in the skip message. FLAKY_FAIL becomes "failed", FLAKY_PASS becomes "flaky_pass",
+// and all other statuses pass through unchanged.
+func mapFlakyStatus(result, message string) string {
+	if result == "skipped" {
+		if strings.HasPrefix(message, "FLAKY_FAIL:") {
+			return "failed"
+		}
+		if strings.HasPrefix(message, "FLAKY_PASS:") {
```
wwared: nit:

```suggestion
		if strings.HasPrefix(message, "FLAKY_PASS") {
```
If there was no reason, there is no `:`, see https://github.com/ethereum-optimism/optimism/pull/19719/changes#diff-445c45812f3bc148965cd8c2091da649cb3a31ff35ae96a52a3cd8e1e5f59c73R261

cci-stats/pkg/service/service.go
```diff
@@ -259,6 +259,21 @@ func fetchJobs(ctx context.Context, client *circleci.Client, workflowID string)
 	return jobs.Items, nil
 }
 
+// mapFlakyStatus maps CircleCI test result status based on flaky markers
+// in the skip message. FLAKY_FAIL becomes "failed", FLAKY_PASS becomes "flaky_pass",
+// and all other statuses pass through unchanged.
+func mapFlakyStatus(result, message string) string {
+	if result == "skipped" {
+		if strings.HasPrefix(message, "FLAKY_FAIL:") {
```
wwared: nit: not necessary, but to keep symmetry with the `FLAKY_PASS` comment above:

```suggestion
		if strings.HasPrefix(message, "FLAKY_FAIL") {
```

cci-stats/pkg/service/service.go
```diff
@@ -283,9 +298,7 @@ func fetchTestMetadata(ctx context.Context, config config.Config, client *circle
 
 	var out []*circleci.TestMetadata
 	for _, m := range md.Items {
-		if m.Result == "skipped" {
-			continue
-		}
+		m.Result = mapFlakyStatus(m.Result, m.Message)
```
wwared, line 301: Just to double check: we want to record every skip and not just the flaky skips? Previously if `m.Result` was skipped we would not append to `out`, but removing the `if` and `continue` block means after mapping to flaky the skipped results get recorded even if they're not flaky skips

EDIT: Nevermind, I see this is the expected intent by re-reading the PR description

wwared (approved):
Seems good, few non-blocking nits and a question below

cci-stats/pkg/service/service.go
ajsutton: Good catch.

cci-stats/pkg/service/service.go
ajsutton: Yeah I'm not sure why we were not recording skips in the first place - it's very useful data to have.

jelias2 (approved):

PR #588, opened 2026-03-27
refactor(proxyd): collapse ConsensusTracker interface and introduce backendState

jelias2:
**Description**
- Replace 6-method ConsensusTracker interface (individual getters/setters) with GetState()/SetState(ConsensusTrackerState) — adding a new consensus field now only requires changing the struct, not the interface
- Introduce backendStateUpdate value struct for setBackendState to eliminate transposition risk with same-typed positional arguments
- ConsensusPoller getters delegate through GetState() rather than individual tracker methods
- postPayload uses GetState() snapshot instead of direct field access

Zero behavior change — all existing consensus tests pass.

**Tests**

**Additional context**

**Metadata**

- Fixes #[Link to Issue]

### DIFF
--- a/proxyd/consensus_poller.go
+++ b/proxyd/consensus_poller.go
@@ -626,7 +626,6 @@ func (cp *ConsensusPoller) Reset() {
 	for _, be := range cp.backendGroup.Backends {
 		cp.backendState[be] = &backendState{}
 	}
-	cp.tracker.SetState(ConsensusTrackerState{})
 }
 
 // fetchBlock is a convenient wrapper to make a request to get a block directly from the backend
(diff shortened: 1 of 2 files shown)


proxyd/consensus_poller.go
```diff
@@ -619,6 +626,7 @@ func (cp *ConsensusPoller) Reset() {
 	for _, be := range cp.backendGroup.Backends {
 		cp.backendState[be] = &backendState{}
 	}
+	cp.tracker.SetState(ConsensusTrackerState{})
```
yashvardhan-kukreja: why is this needed?

jelias2: So looks like the only place we call Reset() is in new consensus poller. I think this actually a great catch, if we are running proxyd in HA mode with a shared backend tracker, this might reset the redis shared state, on a new pod being created which is probably not a desired behavior

jelias2: I'm leaning towards remove it

yashvardhan-kukreja (approved):

PR #590, opened 2026-03-27
refactor(proxyd): extract findConsensusBlock walk-back into reusable function

jelias2:
**Description**

The hash-verification walk-back loop in UpdateBackendGroupConsensus is duplicated by the upcoming CL mode which needs the same logic for safe_l2. Extract it into findConsensusBlock with a blockHashFetcher callback, so CL can reuse it with a different fetcher without duplicating the loop.

No behavior change for EL — elBlockFetcher wraps the existing fetchBlock.

**Tests**

**Additional context**

**Metadata**

- Fixes #[Link to Issue]

### DIFF
--- a/proxyd/consensus_poller.go
+++ b/proxyd/consensus_poller.go
@@ -596,11 +596,10 @@ func (cp *ConsensusPoller) Reset() {
 }
 
 // blockHashFetcher retrieves the block number and hash for a given block from a backend.
-// bs is provided for fetchers that can use cached state; it may be ignored.
-type blockHashFetcher func(ctx context.Context, be *Backend, bs *backendState, block hexutil.Uint64) (hexutil.Uint64, string, error)
+type blockHashFetcher func(ctx context.Context, be *Backend, block hexutil.Uint64) (hexutil.Uint64, string, error)
 
-// elBlockFetcher is a blockHashFetcher for EL backends; bs is unused.
-func (cp *ConsensusPoller) elBlockFetcher(ctx context.Context, be *Backend, _ *backendState, block hexutil.Uint64) (hexutil.Uint64, string, error) {
+// elBlockFetcher is a blockHashFetcher for EL backends.
+func (cp *ConsensusPoller) elBlockFetcher(ctx context.Context, be *Backend, block hexutil.Uint64) (hexutil.Uint64, string, error) {
 	return cp.fetchBlock(ctx, be, block.String())
 }
 
@@ -622,8 +621,8 @@ func (cp *ConsensusPoller) findConsensusBlock(
 
 	for {
 		allAgreed := true
-		for be, bs := range candidates {
-			actualBlockNumber, actualHash, err := fetch(ctx, be, bs, proposedBlock)
+		for be := range candidates {
+			actualBlockNumber, actualHash, err := fetch(ctx, be, proposedBlock)
 			if err != nil {
 				log.Warn("error fetching block for consensus check", "label", label, "name", be.Name, "err", err)
 				continue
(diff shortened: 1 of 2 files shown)


proxyd/consensus_poller.go
```diff
@@ -628,6 +595,65 @@ func (cp *ConsensusPoller) Reset() {
...
+	currentConsensusBlock hexutil.Uint64,
+	startBlock hexutil.Uint64,
+	startHash string,
+	fetch blockHashFetcher,
+	label string,
+) (proposedBlock hexutil.Uint64, proposedBlockHash string, broken bool) {
+	proposedBlock = startBlock
+	proposedBlockHash = startHash
+
+	for {
```
yashvardhan-kukreja, line 622: nit: `for proposedBlock > 0`
Hypothetically a broken backend with even block number 0 could cause proposedBlock -= 1 and make that negative. This would cause errors at line 26 causing a potentially infinite loop due to the `continue` at line 628.

yashvardhan-kukreja: This would also require a `return 0, "", true` outside this loop at the end of this function related, depicting that at least one of the backends didn't agree until block 0.

proxyd/consensus_poller.go
```diff
@@ -484,50 +484,17 @@ func (cp *ConsensusPoller) UpdateBackendGroupConsensus(ctx context.Context) {
...
-			if allAgreed {
-				hasConsensus = true
-			} else {
-				// walk one block behind and try again
-				proposedBlock -= 1
-				proposedBlockHash = ""
-				log.Debug("no consensus, now trying", "block:", proposedBlock)
-			}
-		}
+		proposedBlock, proposedBlockHash, broken = cp.findConsensusBlock(
```
yashvardhan-kukreja, line 496: Subjective and totally up to you: feel free to resolve at your comfort

---

optional nit: the current naming confuses me a little. proposedBlock and currentConsensusBlockNumber have conflicting naming.

I would have loved to see 

```
latestAgreedBlock, latestAgreedBlock, broken = cp.findLastAgreedBlock(
```

proxyd/consensus_poller.go
```diff
@@ -628,6 +595,65 @@ func (cp *ConsensusPoller) Reset() {
 	}
 }
 
+// blockHashFetcher retrieves the block number and hash for a given block from a backend.
+// bs is provided for fetchers that can use cached state; it may be ignored.
+type blockHashFetcher func(ctx context.Context, be *Backend, bs *backendState, block hexutil.Uint64) (hexutil.Uint64, string, error)
```
yashvardhan-kukreja: You should be able to remove `bs *backendState`

If a blockHashFetcher wants to use a backing cache, it can use closures to define it. For example:

```
func (cp *ConsensusPoller) getElBlockFetcher(be *Backend, cache map[hexutil.Uint64]string, block hexutil.Uint64) blockHashFetcher {
    return func(ctx context.Context, be *Backend, block hexutil.Uint64) (hexutil.Uint64, string, error) {
        if cache != nil {
            if val, ok := cache[block]; ok {
            return block, val, nil
            }
        }
        return cp.fetchBlock(ctx, be, block.String())
    }
}

Comment (yashvardhan-kukreja) [reply proxyd/consensus_poller.go]: Actually, given the current implementation it might complicate things a bit.

I'd leave it up to you. Just wanted to share this as an idea.

Personally, it felt like a leaky abstraction when providing a backendState (kinda like interface hinting what the implementation should be)

File: proxyd/consensus_poller.go (hunk 1 above)
Comment (jelias2) [reply proxyd/consensus_poller.go:622]: I've added,
```
....
		if allAgreed {
			return proposedBlock, proposedBlockHash, broken
		}
		if proposedBlock == 0 {
			return 0, "", true
		}
		proposedBlock -= 1
		proposedBlockHash = ""
		log.Debug("no consensus, walking back", "label", label, "block", proposedBlock)
}
```

I agree the not breaking at zero is an issue. 

The proposedBlock is a unsigned int64, so it actually cannot go negative, a decrement at zero will cause a wraparound back to the max uint. 

So adding a check for proposedBlock == 0, after allAgreed is false, would be that the backends do not agree at zero, and we return true for broken with a 0 block and empty hash

jelias2: Adding some test cases too

yashvardhan-kukreja (approved):

PR #592, opened 2026-03-27
refactor(proxyd): rename EL-specific functions and fix type assertion panics

jelias2:
**Description**
Rename generic function names to explicitly EL-scoped names to clarify which path is EL-only and to make the upcoming CL dispatch straightforward to add:
  fetchBlock    → fetchELBlock
  isInSync      → isELInSync
  getPeerCount  → fetchELPeerCount (getPeerCount becomes a thin dispatcher)

Also fix two latent panics in fetchELBlock: previously, missing or non-string number/hash fields in an eth_getBlockByNumber response would panic at the type assertion. They now return errors so the backend is skipped cleanly.

**Tests**

**Additional context**

**Metadata**

- Fixes #[Link to Issue]

### DIFF
--- a/proxyd/consensus_poller.go
+++ b/proxyd/consensus_poller.go
@@ -336,7 +336,7 @@ func (cp *ConsensusPoller) UpdateBackend(ctx context.Context, be *Backend) {
 		return
 	}
 
-	inSync, err := cp.isInSync(ctx, be)
+	inSync, err := cp.isELInSync(ctx, be)
 	RecordConsensusBackendInSync(be, err == nil && inSync)
 	if err != nil {
 		log.Warn("error updating backend sync state", "name", be.Name, "err", err)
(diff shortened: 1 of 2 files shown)


proxyd/consensus_poller.go
```diff
@@ -689,14 +693,27 @@ func (cp *ConsensusPoller) fetchBlock(ctx context.Context, be *Backend, block st
...
+	blockHash = hashStr
 
 	return
 }
 
-// getPeerCount is a convenient wrapper to retrieve the current peer count from the backend
+// getPeerCount retrieves the current peer count from the backend.
 func (cp *ConsensusPoller) getPeerCount(ctx context.Context, be *Backend) (count uint64, err error) {
+	return cp.fetchELPeerCount(ctx, be)
+}
```
jelias2, line 718: getPeerCount will be updated to get CL peer counts in a follow up PR , so fetchELPeerCount gets wrapped inside of generic getPeerCount

yashvardhan-kukreja: > getPeerCount will be updated to get CL peer counts in a follow up PR

I am not sure if that's correct or intended as the current implementation specifically uses an EL-level RPC (net_peerCount).

But let me not block this PR on that. I reckon your subsequent PRs would clarify this for me.

yashvardhan-kukreja (approved):

proxyd/consensus_poller.go
jelias2: Yeah see follow up PR, should make more sense

PR #604, opened 2026-04-15
feat(proxyd): add proxyd_up metric with build info at startup

jelias2:
## Summary
- Adds `proxyd_up{version,commit,date}=1` Prometheus gauge emitted at startup
- Follows standard Prometheus build-info pattern (gauge=1 with metadata in labels)
- Docker builds now inject real git values (`rev-parse HEAD`, `describe --tags --always`, commit timestamp) via `--build-arg` instead of the hardcoded `"docker"` placeholder defaults

## Test plan
- [ ] Run proxyd and verify `proxyd_up` appears in `/metrics` with correct labels
- [ ] Build with `make docker-dev` and confirm labels are populated with real git values

### DIFF
--- a/proxyd/metrics.go
+++ b/proxyd/metrics.go
@@ -31,13 +31,10 @@ var PayloadSizeBuckets = []float64{10, 50, 100, 500, 1000, 5000, 10000, 100000, 
 var MillisecondDurationBuckets = []float64{1, 10, 50, 100, 500, 1000, 5000, 10000, 100000}
 
 var (
-	buildInfo = promauto.NewGaugeVec(prometheus.GaugeOpts{
+	up = promauto.NewGauge(prometheus.GaugeOpts{
 		Namespace: MetricsNamespace,
 		Name:      "up",
-		Help:      "Set to 1 at startup with version info labels.",
-	}, []string{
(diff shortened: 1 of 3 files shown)


proxyd/metrics.go
```diff
@@ -31,6 +31,16 @@ var PayloadSizeBuckets = []float64{10, 50, 100, 500, 1000, 5000, 10000, 100000,
...
 
 var (
+	buildInfo = promauto.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: MetricsNamespace,
+		Name:      "up",
+		Help:      "Set to 1 at startup with version info labels.",
+	}, []string{
+		"version",
+		"commit",
+		"date",
```
sjonpaulbrown: Are there any concerns with the cardinality of these? I can see the value of having `version`, but I am not necessarily seeing the value of including `commit` and `date`. The `version` label would implicitly link to this information.

jelias2: I think commit is valuable for exact traceability I could think of an edge case with duplicate tags, date I could live without. Will remove

sjonpaulbrown (approved):

PR #630, opened 2026-05-21
[codex] test(proxyd): wait for websocket close handler

karlfloersch:
## Summary

Stabilizes `TestWSClientExceedReadLimit` by waiting for the websocket close handler to run instead of checking a boolean immediately after `WriteMessage` returns.

## Why

After #626 fixed CircleCI path filtering and let proxyd tests run again, both stacked proxyd PRs failed in this test at `ws_test.go:239`. The oversized write can return an error before the close handler callback runs, so the old assertion was timing-sensitive.

## Validation

- `go test ./integration_tests -run TestWSClientExceedReadLimit -count=1 -v`

### DIFF
--- a/proxyd/integration_tests/ws_test.go
+++ b/proxyd/integration_tests/ws_test.go
@@ -218,10 +218,13 @@ func TestWSClientExceedReadLimit(t *testing.T) {
 	}, nil)
 	require.NoError(t, err)
 
-	closed := false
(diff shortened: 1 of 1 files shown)


karlfloersch:
Closing this for now. It only addressed a proxyd integration-test flake and is not required for the Optimism dependency bump; leaving that for a separate owner/thread.

PR #640, opened 2026-06-03
feat(proxyd): add agreement interop validation strategy

ajsutton:
Adds an additive `agreement` interop-validation strategy to proxyd. It fans each `checkAccessList` out to all configured supervisor urls concurrently and combines the verdicts by quorum agreement.

- A definitive verdict is valid, or **any** filter JSON-RPC rejection: the supervisor verdict codes (`-3204xx..-3215xx`) *and* the generic invalid-params code (`-32602`, e.g. a malformed/fabricated access list — "failed to parse access entry"). `ParseInteropError` reads the filter's actual JSON-RPC code (including for JSON-RPC errors returned over HTTP 200, which are `rpc.Error` not `rpc.HTTPError`), so rejections are identified by code, not message. Non-verdicts that do not count: failsafe (handled separately), **soft out-of-sync failures (FutureData `-321401` and Uninitialized `-320400`)**, JSON-RPC internal (`-32603`/`-32000`), cancellation (`499`), and transport/5xx.
- **Soft out-of-sync carve-out:** FutureData (`-321401`) and Uninitialized (`-320400`) mean "this node does not have the data yet", not "the tx is invalid". They are treated as non-responses — they never count toward quorum, never short-circuit, and never cause a rejection. A single lagging node is ignored as long as quorum is met by other nodes; if every node is out-of-sync the strategy fails closed (quorum-not-reached).
- **Fast-accept at the minimum:** the strategy decides as soon as `min_responses` definitive verdicts arrive and never awaits slow or unresponded endpoints.
- **Failsafe takes priority among received verdicts:** if any *received* response reports failsafe enabled (dedicated code `-320602`), the message is rejected, even when the minimum accepts are also in hand. A failsafe verdict short-circuits the fan-out; before returning an accept the strategy does a non-blocking drain of already-buffered verdicts so a failsafe that has already arrived beats a met quorum. It never blocks on an endpoint that has not yet responded.
- All-valid (with at least `min_responses` definitive verdicts) accepts; all-invalid rejects with the real error (e.g. a `-32602` parse rejection from every endpoint yields a clean reject, not quorum-not-reached); a mix logs and rejects; fewer than `min_responses` fails closed. Disagreement detection is best-effort (consistent with fast-accept).
- `min_responses` defaults to `len(urls)` (unanimity) and is validated to `1..len(urls)` in `Start`.
- Existing first-supervisor / multicall / health-aware strategies and `performCheckAccessListOp` are unchanged.

Selected via `[interop_validation] strategy = "agreement"` + `min_responses`. Single-endpoint configs are equivalent to today (min defaults to 1).

**Failsafe code source:** failsafe detection keys on the dedicated RPC code `-320602` (`ParseInteropError` routes it independently of message/HTTP status). The op-interop-filter change emitting this dedicated code has landed upstream, so the optimism dependency is bumped and proxyd reads the failsafe, FutureData and Uninitialized codes from `op-core/interop` via `interop.GetErrorCode` rather than hardcoding the literals. Filters must run a build that includes the dedicated failsafe code before this is enabled.

Also includes an unrelated drive-by fix: a pre-existing race in `TestWSClientExceedReadLimit` (the close-handler assertion checked synchronously right after `WriteMessage`, racing the close frame; it failed deterministically on the Gen2 CI runners on `main`). The assertion now polls an `atomic.Bool` via `require.Eventually`; passes 10/10 under `-race`.

This is the proxyd half of a multi-repo change; the op-reth mempool half ships separately.

Test plan: `interop_strategy_test.go` unit tests (failsafe hard-reject by code, failsafe-priority-over-met-quorum, slow non-failsafe endpoint does not delay accept, failsafe short-circuits a slow backend, generic `-32602` parse rejection counts as a definitive invalid, out-of-sync node is soft/ignored, all-out-of-sync fails closed, verdict + soft + failsafe classification coverage) plus `TestInteropValidation_AgreementStrategy` and config tests in `integration_tests/interop_validation_test.go`. Agreement + failsafe + soft tests pass under `-race -count=20`. `gofmt -l .` and `golangci-lint run ./...` are clean. Two unrelated, timing-sensitive integration tests (`TestFrontendMaxRPSLimit`, `TestMaxConcurrentRPCs`) still flake under the race detector and reproduce unchanged on `main`.


proxyd/interop_strategy.go
```diff
@@ -211,6 +212,229 @@ func (s *multicallStrategyImpl) ValidateAccessList(ctx context.Context, interopA
...
+var _ InteropStrategy = (*agreementStrategyImpl)(nil)
+
+func NewAgreementStrategy(urls []string, minResponses int, opts ...commonStrategyOpt) *agreementStrategyImpl {
+	return &agreementStrategyImpl{
+		commonInteropStrategy: NewCommonInteropStrategy(urls, opts...),
+		minResponses:          minResponses,
+	}
+}
+
+func (s *agreementStrategyImpl) ValidateAccessList(ctx context.Context, interopAccessList []common.Hash) error {
```
jelias2, line 255: A nice observability improvement to add might be to time how long this validation takes for an interop transaction. 

If we start finding that waiting for multiple interop-filters responses is taking longer than expected we should be able to observe what type of duration we are adding for transactions to be passed through

proxyd/interop_strategy.go
```diff
@@ -211,6 +212,229 @@ func (s *multicallStrategyImpl) ValidateAccessList(ctx context.Context, interopA
...
+	// Read verdicts as they arrive, breaking once minResponses definitive
+	// verdicts are in. A failsafe verdict short-circuits to a reject. Slow or
+	// unresponded endpoints are never awaited past the quorum.
+	var valid, invalid int
+	var firstInvalid error
+	for range s.urls {
+		v := <-results
+		switch {
+		case v.failsafe:
+			return rejectFailsafe(valid, invalid, v.err)
```
jelias2: Perhaps log which endpoint urls are rejecting the transaction, with error?

Either on 318 or within the reject failsafe function

proxyd/interop_strategy.go
```diff
@@ -211,6 +212,229 @@ func (s *multicallStrategyImpl) ValidateAccessList(ctx context.Context, interopA
...
+		case v.valid:
+			valid++
+		case v.err != nil:
+			invalid++
+			if firstInvalid == nil {
+				firstInvalid = v.err
+			}
+		default:
+			continue // non-response; keep waiting for definitive verdicts
```
jelias2: What exactly would a non-response be in this case? If we have an empty results with nothing in it would that be an unexpected behavior?

Can we add a comment like for readability if this is what would be expected here
```
 transport error, timeout, 5xx, cancel,
```

proxyd/interop_strategy.go
```diff
@@ -211,6 +212,229 @@ func (s *multicallStrategyImpl) ValidateAccessList(ctx context.Context, interopA
...
+// any endpoint is a hard rejection handled before the quorum logic.
+func isFailsafeError(err error) bool {
+	var e *RPCErr
+	return errors.As(err, &e) && e.Code == failsafeInteropRejectionCode
+}
+
+// failsafeInteropRejectionCode is the dedicated supervisor failsafe code. It is
+// handled as a hard short-circuit rejection (see isFailsafeError) and is never a
+// definitive INVALID verdict.
+const failsafeInteropRejectionCode = -320602
```
jelias2: Can there be one source of truth for these codes: failsafeInteropRejectionCode, failsafeInteropRejectionCode, failsafeInteropRejectionCode
It looks like they are defined in backend.go but we are created a new constant for them in this file

jelias2 (approved):
LGTM

PR #642, opened 2026-06-04
feat(proxyd): add metrics for interop validation outcomes

jelias2:
## Summary

Adds a per-transaction Prometheus metric tracking how many interop transactions are **passed through**, **filtered**, or **errored** by interop access list validation.

New counter: `proxyd_rpc_interop_validations_total`, with labels:
- `result` — `passed` | `filtered` | `errored`
- `strategy` — the configured interop validation strategy (e.g. `first-supervisor`, `multicall`, `health-aware-load-balancing`)

It's incremented exactly once per interop transaction in `validateInteropSendRpcRequest`, right after `ValidateAccessList`. This complements the existing per-supervisor-check `rpc_supervisor_checks_total` metric, which counts individual backend checks rather than transaction outcomes.

## Result classification

`interopValidationResult(err)` maps the validation outcome to a label:
- **passed** — `err == nil`
- **filtered** — a client-side `4xx` `*RPCErr` (the supervisor rejected the access list, or it failed a pre-check such as the access-list size limit)
- **errored** — a `5xx` `*RPCErr` or any non-`RPCErr` error (e.g. no supervisor backend available, failsafe enabled, internal errors)

The 4xx/5xx split mirrors how outcomes are already encoded via `http_code` in `rpc_supervisor_checks_total`.

## Testing

- `go build ./...`
- New unit test `TestInteropValidationResult` covering passed/filtered/errored cases (supervisor rejection, access-list-out-of-bounds, failsafe, internal error, backend-unavailable, no-RPC-source).

---

## Rename: `rpc_supervisor_checks_total` → `rpc_interop_filter_checks_total`

Aligns the existing per-check metric with the `InteropFilterClient` naming:

| Before | After |
| --- | --- |
| `rpc_supervisor_checks_total` | `rpc_interop_filter_checks_total` |
| label `supervisor_url` | label `interop_filter_url` |

`http_code`, `rpc_error_code`, and `strategy` labels are unchanged. The matching debug log field was renamed too.

⚠️ **Breaking change** for any dashboard/alert referencing the old metric name or `supervisor_url` label.

PR #643, opened 2026-06-04
feat(proxyd): add metrics for interop validation outcomes

jelias2:
## Summary

Adds a per-transaction Prometheus metric tracking how many interop transactions are **passed through**, **filtered**, or **errored** by interop access list validation, with a breakdown of the specific **reason**.

New counter: `proxyd_rpc_interop_validations_total`, with labels:
- `result` — `passed` | `filtered` | `errored`
- `reason` — the resolved interop RPC error code (or `none`/`internal`); see below
- `strategy` — the configured interop validation strategy

Incremented exactly once per interop transaction in `validateInteropSendRpcRequest`, right after `ValidateAccessList`. This complements the existing per-check supervisor/interop-filter checks metric, which counts individual backend checks rather than transaction outcomes.

## Result classification

`interopValidationResult(err)` maps the validation outcome to a label:
- **passed** — `err == nil`
- **filtered** — a client-side `4xx` `*RPCErr` (the access list was rejected, or it failed a pre-check such as the access-list size limit)
- **errored** — a `5xx` `*RPCErr` or any non-`RPCErr` error (e.g. no backend available, failsafe enabled, internal errors)

## Reason breakdown

`interopValidationReason(err)` **reuses the interop RPC error code** that `ParseInteropError` already resolves — no new mapping is introduced, keeping the diff minimal:
- `none` — passed (`err == nil`)
- the interop RPC error code as a string when rejected — e.g. `-320500` (skipped), `-320600` (conflict), `-32022` (access-list too large), `-320602` (failsafe, dedicated code per optimism#21205), `-32602` (other invalid-params)
- `internal` — a non-`RPCErr` error (e.g. backend unavailable)

Note: failsafe has a dedicated code (`-320602`, optimism#21205), so it is distinguishable from generic invalid-params (`-32602`) automatically. Codes are intentionally **not** redefined in proxyd beyond the existing map entry. A symbolic, human-readable breakdown can be layered on as a follow-up.

## Example metrics

```
# HELP proxyd_rpc_interop_validations_total Count of interop transactions by validation result (passed, filtered, errored) and reason.
# TYPE proxyd_rpc_interop_validations_total counter
proxyd_rpc_interop_validations_total{result="passed",   reason="none",     strategy="multicall"} 142357
proxyd_rpc_interop_validations_total{result="filtered", reason="-320500",  strategy="multicall"} 218
proxyd_rpc_interop_validations_total{result="filtered", reason="-320600",  strategy="multicall"} 17
proxyd_rpc_interop_validations_total{result="filtered", reason="-32022",   strategy="multicall"} 9
proxyd_rpc_interop_validations_total{result="errored",  reason="-320602",  strategy="multicall"} 405
proxyd_rpc_interop_validations_total{result="errored",  reason="internal", strategy="multicall"} 2
```

## Testing

- `go build ./...`
- `TestInteropValidationResult` — passed/filtered/errored classification
- `TestInteropValidationReason` — reason maps to the resolved rpc error code (and `internal` for non-RPCErr)

---
Split out from #642. The supervisor→interop-filter rename is in #644.

### DIFF
--- a/proxyd/interop_validation_result_test.go
+++ b/proxyd/interop_validation_result_test.go
@@ -1,0 +1,59 @@
+package proxyd
+
+import (
+	"errors"
+	"testing"
+
+	interopErrors "github.com/ethereum-optimism/optimism/op-core/interop"
+	"github.com/stretchr/testify/require"
+)
+
+func TestInteropValidationResult(t *testing.T) {
+	tests := []struct {
+		name string
+		err  error
+		want string
+	}{
+		{
+			name: "nil error is passed",
+			err:  nil,
+			want: "passed",
+		},
+		{
+			name: "interop filter rejection (4xx RPCErr) is filtered",
+			err:  interopRPCErrorMap[interopErrors.ErrSkipped], // 422
+			want: "filtered",
+		},
+		{
+			name: "access list out of bounds (413 RPCErr) is filtered",
+			err:  ErrInteropAccessListOutOfBounds, // 413
+			want: "filtered",
+		},
+		{
+			name: "failsafe enabled (503 RPCErr) is errored",
+			err:  interopRPCErrorMap[interopErrors.ErrFailsafeEnabled], // 503
+			want: "errored",
+		},
+		{
+			name: "internal RPCErr (500) is errored",
+			err:  &RPCErr{Code: JSONRPCErrorInternal, HTTPErrorCode: 500},
+			want: "errored",
+		},
+		{
+			name: "non-RPCErr error is errored",
+			err:  errors.New("interop filter backend unavailable"),
+			want: "errored",
+		},
+		{
+			name: "no interop filter backend (non-RPCErr) is errored",
+			err:  interopErrors.ErrNoRPCSource,
+			want: "errored",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			require.Equal(t, tt.want, interopValidationResult(tt.err))
+		})
+	}
+}
--- a/proxyd/interop_validation_result_test.go
+++ b/proxyd/interop_validation_result_test.go
@@ -30,8 +30,18 @@ func TestInteropValidationResult(t *testing.T) {
 			want: "filtered",
 		},
 		{
-			name: "failsafe enabled (503 RPCErr) is errored",
-			err:  interopRPCErrorMap[interopErrors.ErrFailsafeEnabled], // 503
+			name: "failsafe enabled is its own bucket",
+			err:  interopRPCErrorMap[interopErrors.ErrFailsafeEnabled], // -320602
+			want: "failsafe",
+		},
+		{
+			name: "future data (soft out-of-sync, 422) is errored not filtered",
+			err:  interopRPCErrorMap[interopErrors.ErrFuture], // -321401, HTTP 422
+			want: "errored",
+		},
+		{
+			name: "uninitialized (soft out-of-sync, 400) is errored not filtered",
+			err:  interopRPCErrorMap[interopErrors.ErrUninitialized], // -320400, HTTP 400
 			want: "errored",
 		},
 		{
--- a/proxyd/interop_validation_result_test.go
+++ b/proxyd/interop_validation_result_test.go
@@ -85,9 +85,9 @@ func TestInteropValidationReason(t *testing.T) {
 			want: "-32022",
 		},
 		{
-			name: "failsafe maps to its rpc error code",
+			name: "failsafe maps to its dedicated rpc error code",
 			err:  interopRPCErrorMap[interopErrors.ErrFailsafeEnabled],
-			want: "-32602",
+			want: "-320602",
 		},
 		{
 			name: "non-RPCErr is internal",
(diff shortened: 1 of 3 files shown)


proxyd/server.go
```diff
@@ -553,6 +559,47 @@ func (s *Server) validateInteropSendRpcRequest(ctx context.Context, tx *types.Tr
...
+//   - "filtered": the transaction was rejected by interop validation (a client-side
+//     4xx RPCErr, e.g. the interop filter rejected the access list or it failed a pre-check).
+//   - "errored":  validation could not be completed reliably (a 5xx RPCErr or any
+//     non-RPCErr error, e.g. no interop filter backend available or an internal error).
+func interopValidationResult(err error) string {
+	if err == nil {
+		return "passed"
+	}
+	var rpcErr *RPCErr
+	if errors.As(err, &rpcErr) && rpcErr.HTTPErrorCode >= 400 && rpcErr.HTTPErrorCode < 500 {
```
ajsutton: **Claude:** This `4xx → filtered` classification diverges from the codebase's existing `isDefinitiveInteropRejection` (`interop_strategy.go:440`), which deliberately treats several 4xx outcomes as **non-rejections**. Specifically the soft out-of-sync codes `ErrFuture` (`-321401`, HTTP 422) and `ErrUninitialized` (`-320400`, HTTP 400) — excluded via `isSoftInteropFailure` (`interop_strategy.go:416`) — are 4xx, so this labels them `result="filtered"`.

They're reachable as `finalErr`: the **firstSupervisor** strategy returns the backend error directly (`interop_strategy.go:152`), and **multicall** returns `ParseInteropError(firstErr)` with first-error-wins (`interop_strategy.go:213`). So the `filtered` bucket conflates genuine access-list rejections with transient supervisor-lag conditions, and an operator would misread a lagging supervisor as txs being rejected.

Suggest classifying via the existing `isFailsafeError` / `isSoftInteropFailure` / `isDefinitiveInteropRejection` helpers so `filtered` means a definitive INVALID verdict consistently. (The agreement strategy is unaffected — it funnels soft failures into quorum-not-reached → 500 → `errored`.)

proxyd/server.go
```diff
@@ -553,6 +559,47 @@ func (s *Server) validateInteropSendRpcRequest(ctx context.Context, tx *types.Tr
...
 }
 
+// interopValidationResult classifies the outcome of an interop access list
+// validation for metrics purposes:
+//   - "passed":   the transaction was validated successfully (err == nil).
+//   - "filtered": the transaction was rejected by interop validation (a client-side
+//     4xx RPCErr, e.g. the interop filter rejected the access list or it failed a pre-check).
+//   - "errored":  validation could not be completed reliably (a 5xx RPCErr or any
+//     non-RPCErr error, e.g. no interop filter backend available or an internal error).
+func interopValidationResult(err error) string {
```
ajsutton, line 577: **Claude:** (for confirmation, not a blocker) Failsafe (`-320602`, HTTP 503) lands as `result="errored", reason="-320602"`. That's documented and defensible, but failsafe is arguably a deliberate reject-all rather than an error — worth confirming this matches how you'll read dashboards/alerts before merge.

proxyd/interop_validation_result_test.go
```diff
@@ -0,0 +1,109 @@
+package proxyd
+
+import (
+	"errors"
+	"testing"
+
+	interopErrors "github.com/ethereum-optimism/optimism/op-core/interop"
+	"github.com/stretchr/testify/require"
+)
+
```
ajsutton, line 10: **Claude:** Tests don't cover the soft-failure codes (`ErrFuture` -321401 / `ErrUninitialized` -320400) that are exactly the ambiguous case for the `result` classifier — they're 4xx and would currently be labeled `filtered` despite the codebase treating them as non-rejections (see comment on `interopValidationResult`). Whatever classification you land on, please pin it with a test.

ajsutton (approved):
Not sure if we want to differentiate between the soft/definitive/failsafe cases or not as Claude commented below. Approving and leaving it to your judgement.

proxyd/server.go
jelias2: Good catch — fixed in `3892340`. Replaced the raw 4xx-status check with the existing strategy helpers, so `filtered` now means a definitive INVALID verdict via `isDefinitiveInteropRejection` and stays consistent with the validation path. The soft out-of-sync codes (`ErrFuture` -321401, `ErrUninitialized` -320400) are excluded by `isSoftInteropFailure` and now land in `errored` instead of `filtered`, so a lagging supervisor no longer reads as txs being rejected.

proxyd/server.go
jelias2: Thanks for flagging — agreed it's a deliberate reject-all rather than an error. I gave it its own `result="failsafe"` bucket (in `3892340`) instead of folding it into `errored`, so reject-all shows up as a distinct series on dashboards and doesn't get misread as a broken backend. `isDefinitiveInteropRejection` already excludes -320602, so it won't double-count as `filtered`.

proxyd/interop_validation_result_test.go
jelias2: Added in `3892340`: tests now pin the soft codes `ErrFuture` (-321401) and `ErrUninitialized` (-320400) to `errored` (the previously ambiguous 4xx case), plus the failsafe case to the new `failsafe` bucket.

PR #645, opened 2026-06-04
fix(proxyd): allow CL backends at network genesis (block 0 with valid hash)

jelias2:
## Summary

In proxyd's CL consensus mode, `updateCLBackend` rejected any backend reporting block height `0` for **any** of `unsafe`/`safe`/`local_safe`/`finalized`. At network genesis, op-node reports block height `0` for every safety level — but with valid genesis hashes — so every CL backend was excluded and `optimism_syncStatus` never reached clients while the chain was at genesis.

This changes the block-`0` guard to only reject when the corresponding block **hash is also missing/zero** (an uninitialized node). A genesis state (height `0` + valid hash) now passes through; the all-zero initializing case is still rejected.

## Why this is correct (verified against op-node)

`optimism_syncStatus` returns `eth.SyncStatus` (`op-service/eth/sync_status.go`), whose JSON fields match what proxyd parses (`unsafe_l2`, `safe_l2`, `local_safe_l2`, `finalized_l2`, `current_l1`, `head_l1`).

Genesis values come from `currentHeads` in `op-node/rollup/sync/start.go`:

> "If nothing has been marked finalized yet, the finalized head defaults to the genesis block. If nothing has been marked safe yet, the safe head defaults to the finalized block."

So at genesis all L2 safety levels are the genesis block: **number 0 with a valid (non-zero) genesis hash**. The genesis hash is a real keccak hash from the rollup config, always non-zero. Before the first engine reset completes, the status tracker's L2 fields are the zero `L2BlockRef{}` (empty hash) — which `hasValidBlockHash` still rejects, matching the existing `clSyncStatusInitializing` behavior.

## Scope

Scoped to op-node only. op-supernode does **not** serve `optimism_syncStatus` (it exposes `supernode_syncStatus` with a different, nested schema), so this change has no bearing on supernode — supporting it would be a separate change.

The EL path (`consensus_poller.go`) is intentionally left unchanged.

## Changes

- `proxyd/consensus_poller_cl.go`:
  - Capture `SafeBlockHash`/`LocalSafeBlockHash`/`FinalizedBlockHash` in `clSyncStatus` (previously parsed but discarded).
  - Add `hasValidBlockHash` helper.
  - Block-`0` guards now require the hash to also be invalid before rejecting.
- `proxyd/integration_tests/consensus_cl_test.go`:
  - New `TestConsensusCLGenesis` + `clSyncStatusGenesis` helper.

## Testing

- `go test ./proxyd/integration_tests/ -run 'TestConsensusCL'` — pass (incl. existing all-zero "initializing" rejection)
- `go test ./proxyd/...` — pass
- `go vet ./proxyd/...` — clean

### DIFF
--- a/proxyd/consensus_poller_cl.go
+++ b/proxyd/consensus_poller_cl.go
@@ -13,6 +13,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/hexutil"
 	"github.com/ethereum/go-ethereum/log"
 )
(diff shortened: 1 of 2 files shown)


ajsutton (approved):
LGTM.

PR #646, opened 2026-06-04
feat(proxyd): halt CL consensus on unresolvable output root split

jelias2:
**Description**

Adds output-root split protection to proxyd's CL (consensus layer) routing. Each consensus cycle the poller calls `optimism_outputAtBlock` at the group's safe block on every candidate backend and compares the returned output roots:

- **Strict-majority disagreement** — a backend whose root differs from a strict majority of responders is banned (now precisely defined as `topVotes*2 > responders`).
- **No-majority disagreement** — a tie (1v1, 2-2) or an N-way split has no safe winner, so consensus **halts** instead of guessing:
  - the served block height and the cached `optimism_syncStatus` body **freeze** at the last-agreed value, so clients keep seeing the previous response;
  - the served consensus group is **emptied**, so general RPCs **fail closed** (`ErrNoBackends`) rather than being routed to a backend on an uncertain fork.

Backends are **not** persistently banned during a halt — they are still polled and re-verified every cycle — so consensus **auto-recovers** as soon as a majority returns. A halt signals a potential chain split / derivation divergence and warrants operator attention.

This also removes the odd-backend-count startup guard: **any backend count is now permitted**, including even / 2-node cross-client groups (e.g. op-node + kona-node), since ties now halt rather than needing a majority to resolve.

New metrics:
- `proxyd_consensus_cl_halted` — bool gauge, set to 1 while a group is halted (alert target).
- `proxyd_consensus_cl_halt_total` — counter of halted consensus cycles.

**Tests**

- Unit decision tests for `verifyCLOutputRoots`: 1v1 / 3-way / 2-2-1 tie → halt; 2v1 majority bans the minority; unanimous and all-errored → no halt.
- End-to-end `UpdateBackendGroupConsensus` test: a split freezes the served height and empties the group, then auto-recovers (group repopulates, block advances) once backends re-agree.
- CL integration test (`TestConsensusCL`): the all-backends-disagree case now asserts the group fails closed (emptied) with nobody persistently banned.
- All pass under `-race`; the full integration suite is green.

**Additional context**

- The halt is intentionally fail-closed (general RPCs error during a split) rather than serving possibly-forked data — a deliberate safety-over-availability choice for derivation divergence.
- Known edge case: a slow / timing-out **majority** backend can momentarily collapse a 2v1 into a 1v1 tie and trigger a transient (single-cycle) halt that auto-recovers the next cycle. A persistence threshold could be added later if this proves noisy.

**Metadata**

- Fixes #[Link to Issue]

### DIFF
--- a/proxyd/consensus_poller_cl.go
+++ b/proxyd/consensus_poller_cl.go
@@ -50,13 +50,18 @@ func hasValidBlockHash(hash string) bool {
 // derivation position) rather than rewriting individual fields.
 //
 // Output root verification: each poll cycle the poller calls
-// optimism_outputAtBlock at the consensus safe block on every candidate backend
-// and bans any backend whose output root disagrees with the majority. This
-// catches derivation divergences between heterogeneous CL clients (e.g. op-node
+// optimism_outputAtBlock at the consensus safe block on every candidate backend.
+// A backend whose output root disagrees with a strict majority is banned. If there
+// is no majority (a tie or N-way split), the disagreement is unresolvable and
+// consensus halts: the served block height and the cached optimism_syncStatus body
+// stay frozen at the last-agreed value, while the served group is emptied so general
+// RPCs fail closed rather than routing to a backend on an uncertain fork. Backends are
+// not persistently banned, so consensus auto-recovers as soon as a majority returns.
+// This catches derivation divergences between heterogeneous CL clients (e.g. op-node
 // vs kona-node) before incorrect data reaches consumers.
 //
-// An odd number of backends is required (enforced at initialization) so that
-// a majority is always unambiguous.
+// Any number of backends is permitted, including even/2-node cross-client groups:
+// a no-majority disagreement halts rather than being resolved by eviction.
 func WithCLConsensusMode() ConsensusOpt {
 	return func(cp *ConsensusPoller) {
 		cp.consensusLayer = true
@@ -317,7 +322,7 @@ func (cp *ConsensusPoller) fetchCLSyncStatus(ctx context.Context, be *Backend) (
 }
 
 // verifyCLOutputRoots calls optimism_outputAtBlock(safeBlock) on every candidate
-// and bans any backend whose outputRoot disagrees with the majority.
+// and resolves output root agreement before consensus advances.
 //
 // The outputRoot is a cryptographic commitment to the full L2 state at a block:
 //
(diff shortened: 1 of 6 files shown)


proxyd/consensus_poller_cl.go
```diff
@@ -441,34 +453,48 @@ func (cp *ConsensusPoller) verifyCLOutputRoots(
...
+	// Disagreement exists. responders is the number of backends that returned a usable
+	// output root (errors and timeouts excluded), i.e. the total votes cast across all
+	// distinct roots. A backend root is canonical only with a strict majority of the
+	// responders (more than half); otherwise the top is a tie (1v1, 2-2, ...) or an N-way
+	// split, which cannot be resolved safely.
+	responders := 0
+	for _, count := range counts {
+		responders += count
+	}
+	if maxCount*2 <= responders {
```
yashvardhan-kukreja, line 473: TL;DR instead of `if maxCount*2 <= responders`, you should check if the maxCount value occurs only once in the `counts` map (representing non-conflicting majority).

i.e.

```
for root, count := range counts {
    if count == maxCount && root != agreedRoot {
         return 0, 0, halted=true // unresolvable split, coz some other root exists with maxCount number of votes
    }
}
return agreedRoot, halted=false
```

-----

Explanation

Note, previously, the following 

```
{
 "root1": 2,
 "root2": 1,
 "root3": 1,
  "root4": 1
}
```

would have made the function successfully return with root1 as the agreed root.
Because previously:
- maxCount = 2, caused agreedRoot to be returned

But with the newer implementation, the `if maxCount x 2 <= responders` will probably give a failed response as halted=true.

Because in the above example, responders = 5, and maxCount = 2, and hence, maxCount x 2 < responders, causing an unresolvable split as per your code.

jelias2: Let me add a test for this quick

jelias2: Added test: `TestVerifyCLOutputRoots_PluralityNoMajorityHalts` to test this and it does cause a halt which is intended
```
╭─ /private/tmp/proxyd-cl-build/proxyd jelias2/cl-halt-on-tie *4                                                                                                                                                                                                                                                      10:50:37
╰─❯ go test -run 'TestVerifyCLOutputRoots_(Tie|ThreeWaySplit|TopTie|Plurality|Majority|Unanimous|AllErrored|Timeouts)' -v ./
=== RUN   TestVerifyCLOutputRoots_TieHalts
--- PASS: TestVerifyCLOutputRoots_TieHalts (0.00s)
=== RUN   TestVerifyCLOutputRoots_ThreeWaySplitHalts
--- PASS: TestVerifyCLOutputRoots_ThreeWaySplitHalts (0.00s)
=== RUN   TestVerifyCLOutputRoots_TopTieHalts
--- PASS: TestVerifyCLOutputRoots_TopTieHalts (0.00s)
=== RUN   TestVerifyCLOutputRoots_PluralityNoMajorityHalts
--- PASS: TestVerifyCLOutputRoots_PluralityNoMajorityHalts (0.00s)
=== RUN   TestVerifyCLOutputRoots_MajorityResolvesNoHalt
--- PASS: TestVerifyCLOutputRoots_MajorityResolvesNoHalt (0.00s)
=== RUN   TestVerifyCLOutputRoots_MajorityOverMultipleMinoritiesResolves
--- PASS: TestVerifyCLOutputRoots_MajorityOverMultipleMinoritiesResolves (0.00s)
=== RUN   TestVerifyCLOutputRoots_UnanimousNoHalt
--- PASS: TestVerifyCLOutputRoots_UnanimousNoHalt (0.00s)
=== RUN   TestVerifyCLOutputRoots_AllErroredNoHalt
--- PASS: TestVerifyCLOutputRoots_AllErroredNoHalt (0.50s)
=== RUN   TestVerifyCLOutputRoots_TimeoutsExcludedFromResponderCount
--- PASS: TestVerifyCLOutputRoots_TimeoutsExcludedFromResponderCount (0.50s)
PASS
```

yashvardhan-kukreja: Update: we agreed on the newer approach favoring a stronger, despite more eventually consistency. TL;DR the newer approach would only agree on a root if has at least 51% of votes (i.e. maxVotes * 2 > participants)

yashvardhan-kukreja: Update: we agreed on the newer approach favoring a stronger, despite more eventually consistency. TL;DR the newer approach would only agree on a root if has at least 51% of votes (i.e. maxVotes * 2 > participants)

yashvardhan-kukreja (approved):

PR #648, opened 2026-06-09
refactor(proxyd): rename supervisor terminology to interop-filter

ajsutton:
Pre-production rename of all supervisor → interop-filter terminology: code, the `first-interop-filter` strategy value, the `rpc_interop_filter_checks_total` metric and `interop_filter_url` label, comments, and tests. No behavior change.


karlfloersch (approved):
🚢

jelias2:
Can if its not too much work can this make the config flags backwards compatible or else we would have cut a major bump to v5 with a breaking change

ajsutton:
@jelias2 Made the config flag work with old and new names.

jelias2 (approved):

PR #654, opened 2026-06-09
refactor(proxyd): unified transaction filter on the submission path

ajsutton:
Replaces the method-string `if` branches in `handleBatchRPC` with a single `TxFilter` chokepoint that decodes each submission once and runs an ordered module pipeline: **ChainID → SenderRateLimit → Interop → TxMiddleware**, short-circuiting on the first reject. Sender recovery is memoized in `TxSubmission.Sender` and shared across modules.

The four modules (`proxyd/tx_filter_modules.go`) each wrap logic that existed before; the interop module inlines the interop sender rate-limit (no chain-ID re-check, since chain-ID is now its own module) and preserves the `rpcInteropValidationsTotal` metric. The configurable `tx_validation.methods` gate is retained on the middleware module. The `maxBundleTransactions` cap moves into `TxFilter.Build` so it always runs.

Closes the bundle→interop gap: every tx in an `eth_sendBundle` now flows through the same filter as single-tx submissions.

### Behavior changes (intentional)
- **B1** — Bundles now get interop validation, and are decoded + size-capped even when the middleware is disabled. A malformed bundle that previously passed undecoded is now rejected at proxyd; the bundle-size cap applies to every bundle.
- **B2** — Bundles now get the chain-ID check and per-`sender:nonce` rate-limit. Per-tx quota consumed before a later tx rejects is not rolled back (same model as the prior single-tx path).
- **B3** — Chain-ID enforcement is decoupled from the rate limiter: the ChainID module is active whenever `allowed_chain_ids` is non-empty, independent of either rate limiter.

Deleted: the two inline `handleBatchRPC` blocks, `applyTxValidation`, `validateInteropSendRpcRequest`, `rateLimitSender`, `genericRateLimitSender`, `rateLimitInteropSender` (as a \*Server method), and the dead `SetTxValidationFn`. `isAllowedChainId` is now a package function used by the ChainID module.

Plan: `op-claude/plans/2026-06-09-refactor-proxyd-unified-tx-filter-plan.md`.

### Test plan
- `go build ./...`, `golangci-lint run ./...` (0 issues), `go test ./...` (unit + integration) all pass.
- New unit tests cover filter ordering/short-circuit, decode errors, the bundle-size cap, sender memoization/recovery-error, and each module.
- New integration tests (`tx_filter_test.go`): bundle→interop enforcement (incl. middleware-disabled + cap), bundle chain-ID + per-tx sender rate-limit, order short-circuit, chain-ID enforced with rate limiters off, and middleware method-config gating.

### DIFF
--- a/proxyd/server.go
+++ b/proxyd/server.go
@@ -21,7 +21,6 @@ import (
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/hexutil"
-	"github.com/ethereum/go-ethereum/core/txpool"
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/core/types/interoptypes"
 	"github.com/ethereum/go-ethereum/log"
@@ -98,6 +97,7 @@ type Server struct {
 	txValidationMethods      TxValidationMethodSet
 	txValidationClient       *TxValidationClient
 	txValidationFailOpen     bool
+	txFilter                 *TxFilter
 	isDraining               atomic.Bool
 	gracefulShutdownDuration time.Duration
 }
@@ -221,7 +221,7 @@ func NewServer(
 		txValidationFailOpen = *txValidationConfig.FailOpen
 	}
 
-	return &Server{
+	srv := &Server{
 		BackendGroups:        backendGroups,
 		wsBackendGroup:       wsBackendGroup,
 		wsMethodWhitelist:    wsMethodWhitelist,
@@ -259,7 +259,31 @@ func NewServer(
 		txValidationClient:       txValidationClient,
 		txValidationFailOpen:     txValidationFailOpen,
 		gracefulShutdownDuration: gracefulShutdownDuration,
-	}, nil
+	}
+
+	var modules []TxFilterModule
+	if len(srv.allowedChainIds) > 0 { // B3: independent of the rate limiter
+		modules = append(modules, &chainIDModule{allowedChainIds: srv.allowedChainIds})
+	}
+	if srv.senderLim != nil {
+		modules = append(modules, &senderRateLimitModule{lim: srv.senderLim})
+	}
+	modules = append(modules, &interopModule{ // strategy always set post-#649
+		strategy:         srv.interopStrategy,
+		interopSenderLim: srv.interopSenderLim,
+		validatingCfg:    srv.interopValidatingConfig,
+	})
+	if srv.enableTxValidation {
+		modules = append(modules, &txMiddlewareModule{
+			endpoint: srv.txValidationEndpoint,
+			fn:       srv.txValidationFn,
+			failOpen: srv.txValidationFailOpen,
+			methods:  srv.txValidationMethods,
+		})
+	}
+	srv.txFilter = NewTxFilter(srv, modules...)
+
+	return srv, nil
 }
 
 func (s *Server) RPCListenAndServe(host string, port int) error {
@@ -520,45 +544,6 @@ func checkInteropAndReturnAccessList(ctx context.Context, tx *types.Transaction)
 	return interopAccessList, true, nil
 }
 
-func (s *Server) validateInteropSendRpcRequest(ctx context.Context, tx *types.Transaction) error {
-	interopAccessList, isInterop, err := checkInteropAndReturnAccessList(ctx, tx)
-	if err != nil {
-		return err
-	}
-	if !isInterop {
-		return nil
-	}
-	// at this point, we know it's an interop transaction worthy of being validated
-	log.Info(
-		"validating interop access list",
-		"source", "rpc",
-		"req_id", GetReqID(ctx),
-		"strategy", s.interopValidatingConfig.Strategy,
-		"tx_hash", tx.Hash(),
-	)
-	if err := s.rateLimitInteropSender(ctx, tx); err != nil {
-		return err
-	}
-	if err := reqSizeLimitCheck(ctx, tx, s.interopValidatingConfig.ReqSizeLimit); err != nil {
-		return err
-	}
-
-	finalErr := s.interopStrategy.ValidateAccessList(ctx, interopAccessList)
-
-	rpcInteropValidationsTotal.WithLabelValues(
-		interopValidationResult(finalErr),
-		interopValidationReason(finalErr),
-		string(s.interopValidatingConfig.Strategy),
-	).Inc()
-
-	if finalErr == nil {
-		log.Info("interop access list validated successfully", "req_id", GetReqID(ctx), "tx_hash", tx.Hash())
-	} else {
-		log.Info("interop access list validation failed", "req_id", GetReqID(ctx), "tx_hash", tx.Hash(), "error", finalErr)
-	}
-	return finalErr
-}
-
 // interopValidationResult classifies the outcome of an interop access list
 // validation for metrics purposes. It keys on the same helpers the validation
 // strategies use, rather than the raw HTTP status, so the buckets match how the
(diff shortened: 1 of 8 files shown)


proxyd/tx_filter.go
```diff
@@ -0,0 +1,143 @@
...
+		return nil, ErrInvalidParams(fmt.Sprintf("bundle contains %d transactions, maximum allowed is %d", len(txs), maxBundleTransactions))
+	}
+
+	return &TxSubmission{
+		Method:          req.Method,
+		Req:             req,
+		Txs:             txs,
+		BypassRateLimit: bypassRateLimit,
+	}, nil
+}
```
jelias2, line 145: **Future-proof the submission-method set: one declarative table, no drift.**

Right now the method set lives in two `switch` statements — here in `IsSubmissionMethod` and again in `Build` — that can silently diverge. Since this set gates chain-ID, rate-limit, and interop, a method that's in one switch but not the other is a filter gap. Collapse both into a single table keyed by method, carrying each method's decoder, so adding a method is one edit and the two can't drift:

```suggestion
// submissionDecoder decodes a submission request into its transactions. One
// entry per supported method is the single source of truth: membership
// (IsSubmissionMethod) and decoding (Build) both read this map, so they cannot
// drift, and adding a method is a single edit guarded by a pinning test.
type submissionDecoder func(ctx context.Context, f *TxFilter, req *RPCReq) ([]*types.Transaction, error)

var submissionMethods = map[string]submissionDecoder{
	"eth_sendRawTransaction":            decodeSingleTx,
	"eth_sendRawTransactionConditional": decodeSingleTx,
	"eth_sendBundle":                    decodeBundle,
}

func decodeSingleTx(ctx context.Context, f *TxFilter, req *RPCReq) ([]*types.Transaction, error) {
	tx, err := f.decode(ctx, req)
	if err != nil {
		return nil, err
	}
	return []*types.Transaction{tx}, nil
}

func decodeBundle(ctx context.Context, _ *TxFilter, req *RPCReq) ([]*types.Transaction, error) {
	return transactionsFromBundleReq(ctx, req)
}

// IsSubmissionMethod reports the static set of submission methods the filter
// handles. Per-method middleware opt-out is enforced inside the middleware
// module, not here — chain-ID/rate-limit/interop always apply.
func (f *TxFilter) IsSubmissionMethod(method string) bool {
	_, ok := submissionMethods[method]
	return ok
}

// Build decodes a submission request into a TxSubmission using the decoder
// registered for its method, and enforces maxBundleTransactions here so the cap
// applies to every bundle regardless of middleware enablement.
func (f *TxFilter) Build(ctx context.Context, req *RPCReq, bypassRateLimit bool) (*TxSubmission, error) {
	decode, ok := submissionMethods[req.Method]
	if !ok {
		return nil, fmt.Errorf("not a submission method: %s", req.Method)
	}
	txs, err := decode(ctx, f, req)
	if err != nil {
		return nil, err
	}

	if len(txs) > maxBundleTransactions {
		log.Warn("bundle contains too many transactions",
			"req_id", GetReqID(ctx),
			"tx_count", len(txs),
			"max_allowed", maxBundleTransactions)
		return nil, ErrInvalidParams(fmt.Sprintf("bundle contains %d transactions, maximum allowed is %d", len(txs), maxBundleTransactions))
	}

	return &TxSubmission{
		Method:          req.Method,
		Req:             req,
		Txs:             txs,
		BypassRateLimit: bypassRateLimit,
	}, nil
}
```

Pair it with a pinning test in `tx_filter_test.go` so the set can only change deliberately (this is what turns a silent omission into a reviewed decision):

```go
func TestSubmissionMethods_Pinned(t *testing.T) {
	got := make([]string, 0, len(submissionMethods))
	for m := range submissionMethods {
		got = append(got, m)
	}
	sort.Strings(got)
	require.Equal(t, []string{
		"eth_sendBundle",
		"eth_sendRawTransaction",
		"eth_sendRawTransactionConditional",
	}, got, "changing the submission-method set gates chain-ID, rate-limit, and "+
		"interop — update this test deliberately and confirm the new method decodes correctly")
}
```

proxyd/tx_filter_modules.go
```diff
@@ -0,0 +1,158 @@
...
+			interopValidationResult(finalErr),
+			interopValidationReason(finalErr),
+			string(m.validatingCfg.Strategy),
+		).Inc()
+
+		if finalErr == nil {
+			log.Info("interop access list validated successfully", "req_id", GetReqID(ctx), "tx_hash", tx.Hash())
+		} else {
+			log.Info("interop access list validation failed", "req_id", GetReqID(ctx), "tx_hash", tx.Hash(), "error", finalErr)
+			return finalErr
```
jelias2, line 135: **Add a test for the bundle short-circuit here.** This `return finalErr` is the load-bearing line for bundles: the first interop tx to fail must reject the whole submission and stop validation — no later tx should be validated. There's currently no test asserting that across a multi-tx bundle.

Because the `rpcInteropValidationsTotal` increment is co-located with each `ValidateAccessList` call (just above), the strategy's call count also equals the number of metric increments — so asserting the call count proves both the short-circuit *and* that only the txs actually validated (up to and including the rejecter) are counted, never the short-circuited tail.

The existing `fakeInteropStrategy` returns a fixed result per call, so add a tiny call-indexed strategy and a test in `tx_filter_modules_test.go`:

```go
// failOnNthStrategy rejects the Nth interop tx it validates (1-based) and
// accepts the others, counting calls.
type failOnNthStrategy struct {
	failCall int
	calls    int
}

func (s *failOnNthStrategy) ValidateAccessList(ctx context.Context, _ []common.Hash) error {
	s.calls++
	if s.calls == s.failCall {
		return ErrInternal
	}
	return nil
}

// TestInteropModule_BundleShortCircuit proves that within a bundle the first
// failing interop tx rejects the whole submission and no later tx is validated.
func TestInteropModule_BundleShortCircuit(t *testing.T) {
	strategy := &failOnNthStrategy{failCall: 2}
	m := &interopModule{strategy: strategy, validatingCfg: InteropValidationConfig{}}
	sub := &TxSubmission{Txs: []*types.Transaction{
		interopTx(t, big.NewInt(1), 0), // call 1: passes
		interopTx(t, big.NewInt(1), 1), // call 2: fails -> reject
		interopTx(t, big.NewInt(1), 2), // never validated
	}}

	require.ErrorIs(t, m.Apply(context.Background(), sub), ErrInternal)
	require.Equal(t, 2, strategy.calls,
		"validation must stop at the first failing tx; the 3rd interop tx (and its metric increment) must never happen")
}
```

If you want to assert the metric directly rather than via the co-located call count, `prometheus/client_golang/prometheus/testutil.CollectAndCount` on `rpcInteropValidationsTotal` works too, but it needs a counter reset for isolation — the call-count assertion above is the simpler, equivalent check.

jelias2 (commented):
Follow-ups from review: (1) collapse the submission-method set into one declarative table to prevent drift + pinning test, (2) test the interop bundle short-circuit, (3) fence the WS-whitelist bypass at startup (suggested as follow-up).

proxyd/server.go
```diff
@@ -259,7 +259,38 @@ func NewServer(
...
+// and is the single source of truth for module sequencing.
+func (s *Server) txFilterModules() []TxFilterModule {
+	var modules []TxFilterModule
+	if len(s.allowedChainIds) > 0 { // chain-ID enforcement is independent of either rate limiter
+		modules = append(modules, &chainIDModule{allowedChainIds: s.allowedChainIds})
+	}
+	if s.senderLim != nil {
+		modules = append(modules, &senderRateLimitModule{lim: s.senderLim})
+	}
+	modules = append(modules, &interopModule{ // interopStrategy is always non-nil
```
jelias2, line 280: **Enforce the "always non-nil" invariant instead of just asserting it in a comment.** `interopModule` calls `m.strategy.ValidateAccessList(...)` (`tx_filter_modules.go:123`) with no nil check, relying on this comment. It's true today — `proxyd.go` always resolves a strategy (defaults empty → first-filter, errors on unknown) — so there's no live crash path. But the guarantee lives only in the caller, and a future caller or test that constructs a `Server`/`interopModule` without a strategy gets a nil-pointer panic with no actionable error.

Cheap to make the invariant self-enforcing — guard at the top of `NewServer`:

```go
if interopStrategy == nil {
	return nil, fmt.Errorf("interopStrategy must not be nil")
}
```

Not a blocker (no reachable nil path in production), but it turns a latent panic into a clear startup error.

proxyd/tx_filter.go
```diff
@@ -0,0 +1,143 @@
...
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/log"
+)
+
+// TxSubmission is the decoded form of a single transaction-submission request
+// (eth_sendRawTransaction, eth_sendRawTransactionConditional, eth_sendBundle).
+// The submission is decoded once and the recovered sender for each tx is
+// memoized so that the chain-ID, sender-rate-limit, and interop modules can all
+// share a single ecrecover per transaction.
```
jelias2, line 16: **Document the single-goroutine contract on the type.** `Sender` lazily initializes the `senders`/`senderErr`/`recovered` slices (nil-check then allocate), which is racy if two goroutines call it on the same `*TxSubmission`. It's safe today — `handleBatchRPC` builds one `TxSubmission` per request and `TxFilter.Apply` runs modules sequentially (`-race` is clean) — but nothing on the type says so, and the existing doc only covers `sub.Txs` immutability. Make the contract explicit so a future concurrent caller doesn't trip over it:

```suggestion
// TxSubmission is the decoded form of a single transaction-submission request
// (eth_sendRawTransaction, eth_sendRawTransactionConditional, eth_sendBundle).
// The submission is decoded once and the recovered sender for each tx is
// memoized so that the chain-ID, sender-rate-limit, and interop modules can all
// share a single ecrecover per transaction.
//
// TxSubmission is NOT safe for concurrent use: Sender memoizes via lazy slice
// init. The filter pipeline runs modules sequentially over a per-request
// submission, so this is sufficient — do not share a TxSubmission across goroutines.
```

jelias2 (commented):
Two more from the Go review: enforce the interopStrategy non-nil invariant (H1) and document TxSubmission's single-goroutine contract (M2).

proxyd/server.go
ajsutton: Passing nil for a required field is a programming error so panic is the right result. The process should fail hard and loud. Claude really likes to try to limp along with these "someone might stupid in the future" checks but there's a bit of an anti-pattern.

jelias2 (approved):

PR #655, opened 2026-06-11
fix(proxyd): route websocket txs through filters

ajsutton:
Routes websocket transaction RPCs through the same pre-forward validation and backend-group forwarding path used by HTTP. This keeps subscription-style websocket traffic on the pinned upstream connection, while mapped RPC methods now get sender limits, interop validation, tx validation middleware, and HTTP-style backend selection.

Also enforces `max_ws_conns` with a per-backend websocket semaphore and records WS-routed backend forwards with the WS source label.

Additional coverage locks in that whitelisted-but-unmapped transaction forwarding methods fail closed locally, while whitelisted subscription methods that are not RPC-mapped still fall through to the WS upstream.

Test plan:
- `go test ./integration_tests -run 'TestWS(MappedTransactionUsesHTTPPipeline|WhitelistedUnmappedTransactionRejected|WhitelistedUnmappedSubscriptionUsesWSBackend)$' -count=1 -v`
- `go test .`
- `go test ./integration_tests -count=1`

### DIFF
--- a/proxyd/backend.go
+++ b/proxyd/backend.go
@@ -1444,7 +1444,9 @@ type WSProxier struct {
 	writeTimeout     time.Duration
 	onClose          func()
 	closeOnce        sync.Once
-	requestProcessor func(context.Context, *RPCReq) (*RPCRes, bool)
+	requestHandler   func(*RPCReq) bool
+	requestProcessor func(context.Context, *RPCReq) *RPCRes
+	requestSem       *semaphore.Weighted
 }
 
 func NewWSProxier(backend *Backend, clientConn, backendConn *websocket.Conn, methodWhitelist *StringSet, onClose func()) *WSProxier {
@@ -1456,6 +1458,7 @@ func NewWSProxier(backend *Backend, clientConn, backendConn *websocket.Conn, met
 		readTimeout:     defaultWSReadTimeout,
 		writeTimeout:    defaultWSWriteTimeout,
 		onClose:         onClose,
+		requestSem:      semaphore.NewWeighted(defaultMaxConcurrentWSRPCs),
 	}
 }
 
@@ -1535,10 +1538,10 @@ func (w *WSProxier) clientPump(ctx context.Context, errC chan error) {
 			continue
 		}
 
-		if w.requestProcessor != nil {
-			res, handled := w.requestProcessor(ctx, req)
-			if handled {
-				msg = mustMarshalJSON(res)
+		if w.requestHandler != nil && w.requestHandler(req) {
+			if !w.acquireRequestSlot() {
+				msg = mustMarshalJSON(NewRPCErrorRes(req.ID, ErrTooManyRequests))
+				RecordRPCError(ctx, BackendProxyd, req.Method, ErrTooManyRequests)
 				err = w.writeClientConn(msgType, msg)
 				if err != nil {
 					errC <- err
@@ -1546,6 +1549,8 @@ func (w *WSProxier) clientPump(ctx context.Context, errC chan error) {
 				}
 				continue
 			}
+			go w.processClientRPC(ctx, msgType, req, errC)
+			continue
 		}
 
 		RecordRPCForward(ctx, w.backend.Name, req.Method, RPCRequestSourceWS)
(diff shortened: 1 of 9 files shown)


proxyd/backend.go
```diff
@@ -1525,6 +1569,46 @@ func (w *WSProxier) clientPump(ctx context.Context, errC chan error) {
...
+	return w.requestSem.TryAcquire(1)
+}
+
+func (w *WSProxier) releaseRequestSlot() {
+	if w.requestSem != nil {
+		w.requestSem.Release(1)
+	}
+}
+
+func (w *WSProxier) processClientRPC(ctx context.Context, msgType int, req *RPCReq, errC chan error) {
```
jelias2, line 1595: processClientRPC goroutines have no cancellation (backend.go:1465–1471, 1552).
  
Proxy reads one error from errC, calls close(), returns. In-flight processClientRPC goroutines (up to defaultMaxConcurrentWSRPCs=100/conn) keep running on the context.WithoutCancel ctx — through the full retry/backoff loop — against a torn-down connection. Bounded, so not unbounded leak, but they can linger seconds per request.

 Suggestion: thread a cancelable context into the proxier, cancel it in close().

proxyd/server.go
```diff
@@ -46,6 +46,7 @@ const (
 	defaultWSHandshakeTimeout                       = 10 * time.Second
 	defaultWSReadTimeout                            = 2 * time.Minute
 	defaultWSWriteTimeout                           = 10 * time.Second
+	defaultMaxConcurrentWSRPCs                      = 100
```
jelias2, line 49: Should this be a configurable value instead of hardcoded?

proxyd/backend.go
```diff
@@ -733,8 +739,16 @@ func (b *Backend) Forward(ctx context.Context, reqs []*RPCReq, isBatch bool) ([]
 }
 
 func (b *Backend) ProxyWS(clientConn *websocket.Conn, methodWhitelist *StringSet, serverReadLimit int64) (*WSProxier, error) {
+	if b.wsSem != nil && !b.wsSem.TryAcquire(1) {
+		return nil, ErrBackendOverCapacity
+	}
```
jelias2, line 744: This seems to fail silently consider warn logging if error occurs here

jelias2 (commented):
Couple Testing Gaps

  1. Per-backend max_ws_conns enforcement (the headline change). Nothing exercises wsSem. Missing:
  - Nth+1 WS connection to a backend at capacity is rejected / falls over to the next backend (ErrBackendOverCapacity skip in BackendGroup.ProxyWS, backend.go:1332).
  - A slot is released on disconnect (onClose → wsSem.Release), so a new connection succeeds after one closes. This is the bug-prone part — if release breaks, you leak capacity until restart.
  - Release-on-dial-failure (backend.go:750) — acquired slot is freed when the backend dial fails.

  TestWSProxierRequestSlotLimit tests the per-request requestSem helper in isolation; it does not touch the per-connection wsSem.

  2. requestSem exhaustion end-to-end. No test that the 101st concurrent mapped RPC on one connection gets an ErrTooManyRequests JSON response over the wire. Only the in-process helper is tested.

  Worth adding — error branches in handleWSRPC

  - Backend forward returns an error → client gets the error response (server.go:889).
  - Backend returns empty res → ErrInternal (server.go:896). Pure defensive branch, currently 0% covered.

  Coverage gaps in routing conditions

  - shouldHandleWSRPC conditions 2/3 are only tested via eth_sendRawTransaction (a submission method). No test for a tx-validation-only method (enableTxValidation && txValidationMethods.Contains) being intercepted over WS.
  - An actual interop tx over WS going through the filter — the submission path is tested but not the interop module engaging on the WS path.

proxyd/server.go
```diff
@@ -952,6 +1002,50 @@ func (s *Server) isValidAPIKey(key string) bool {
 	return false
 }
 
+func (s *Server) limiterForRequest(ctx context.Context, r *http.Request, bypassLimit bool) (limiterFunc, string, string, string, error) {
```
jelias2: Worth doing — limiterForRequest over-returns

  It returns (limiterFunc, origin, userAgent, xff, error) but:
  - HandleWS discards three: isLimited, _, _, _, err (server.go:819).
  - HandleRPC uses origin/userAgent/xff only for one log.Debug line (server.go:387–394).

  The three strings are leaking the function's internals to satisfy one debug log. Collapse to (limiterFunc, error) and let HandleRPC read r.Header.Get("Origin") / stripXFF(GetXForwardedFor(ctx)) inline for its log. 5-value return → 2, and the WS caller stops pretending to care about return values it throws away.

jelias2 (commented):
overall LGTM couple nits

jelias2 (approved):

PR #663, opened 2026-06-18
proxyd: Fix a race between reading results and draining discarded ones in multi call strategy

ajsutton:
The drain isn't required as the channel is big enough to hold all responses anyway. Previously there would be a race between the two readers. This could result in a deadlock if all responses returned an error causing the result reader to actually read all responses if the last one was actually read by the draining thread that was unblocked when the last response was _written_ (but possibly not yet read).

### DIFF
--- a/proxyd/Makefile
+++ b/proxyd/Makefile
@@ -25,6 +25,10 @@ test:
 	go test -v ./...
 .PHONY: test
 
+test-race:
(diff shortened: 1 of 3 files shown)


karlfloersch (approved):
Good change given we don't need the wg

LGTM

jelias2 (approved):

PR #664, opened 2026-06-24
ci: add beta parallel apko image builds

falcorocks:
## Summary

- add a parallel apko image build workflow while keeping the existing Docker-based image workflow unchanged
- add melange and apko configuration for the covered images
- publish apko images to a separate image namespace to avoid collisions with the existing image outputs
- keep tag behavior aligned with the existing image build flow

## Vulnerability comparison

Measured with `grype 0.107.0` against the Docker images from `main` (`a82c8a58b988872a4e6dd34d06a1c4c9b1c0f75b`) and the apko images from this PR's workflow (`8f273af1bd93b8e8e6915c8359553e2e3a28b453`). Both amd64 and arm64 were measured where a Docker baseline exists; counts were identical across platforms, so the table is collapsed per image.

Totals include all severities reported by Grype. `C/H/M/L` is Critical/High/Medium/Low.

| Image | Docker total | apko total | Change | Reduction | Docker C/H/M/L | apko C/H/M/L |
|---|---:|---:|---:|---:|---:|---:|
| `cci-stats` | 6 | 6 | 0 | 0.0% | 2/1/2/1 | 2/1/2/1 |
| `circleci-runner-wolfi` | n/a | 1 | n/a | n/a | n/a | 0/0/1/0 |
| `op-conductor-mon` | 50 | 3 | -47 | 94.0% | 1/24/23/2 | 0/1/2/0 |
| `op-conductor-ops` | 41 | 14 | -27 | 65.9% | 0/10/17/3 | 0/5/7/1 |
| `op-signer` | 6 | 6 | 0 | 0.0% | 1/2/3/0 | 1/2/3/0 |
| `op-txproxy` | 6 | 6 | 0 | 0.0% | 1/3/2/0 | 1/3/2/0 |
| `op-ufm` | 7 | 7 | 0 | 0.0% | 1/2/4/0 | 1/2/4/0 |
| `peer-mgmt-service` | 14 | 14 | 0 | 0.0% | 0/4/10/0 | 0/4/10/0 |
| `proxyd` | 44 | 2 | -42 | 95.5% | 1/24/19/0 | 0/1/1/0 |
| `replica-healthcheck` | 71 | 61 | -10 | 14.1% | 0/28/29/14 | 0/25/25/11 |

### DIFF
--- a/.github/images.apko.json
+++ b/.github/images.apko.json
@@ -1,0 +1,168 @@
+{
+  "shared_paths": [
+    ".github/images.apko.json",
+    ".github/workflows/build-images.apko.yaml",
+    "melange/pipelines/**",
+    "go.mod"
+  ],
+  "melange_archs": [
+    "x86_64",
+    "aarch64"
+  ],
+  "apko_archs": "amd64,arm64",
+  "release": {
+    "tag_regex": "^([^/]+)/(v.+)$",
+    "service_match": 1,
+    "version_match": 2,
+    "publish_tag": "${version}",
+    "source_ref": "${ref_name}"
+  },
+  "default_runners": {
+    "x86_64": "ubuntu-24.04",
+    "aarch64": "ubuntu-24.04-arm"
+  },
+  "melange": {
+    "cci-stats": {},
+    "op-conductor-mon": {},
+    "op-signer": {},
+    "op-txproxy": {},
+    "op-ufm": {},
+    "peer-mgmt-service": {},
+    "proxyd": {},
+    "op-conductor-ops": {},
+    "replica-healthcheck": {},
+    "circleci-runner": {}
+  },
+  "images": {
+    "cci-stats": {
+      "needs_melange": [
+        "cci-stats"
+      ],
+      "paths": [
+        "cci-stats/**"
+      ],
+      "nexus_paths": [
+        "melange/cci-stats.yaml",
+        "apko/cci-stats.yaml"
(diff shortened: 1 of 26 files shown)


raffaele-oplabs (approved):
Approving because the APKO workflow is additive and does not change the current Docker publishing path.

Future follow-ups before treating these APKO images as drop-in replacements:

- `proxyd` should carry over the Docker image default config argument. The Docker image has `CMD ["/etc/proxyd/proxyd.toml"]`; the APKO image currently only sets the entrypoint. Deployments that rely on the image default would start without that config path.
- `op-ufm` has the same issue: the Docker image has `CMD ["/etc/ufm/config.toml"]`, while the APKO image only sets the entrypoint.
- The smoke tests use `--help`, so they are unlikely to catch missing default runtime args. Add a startup/config smoke check before promoting these images.

sjonpaulbrown (approved):
This is for consideration. I don't see it as blocking, but it would be worth reviewing as I am not sure it ever acctually worked as intended. We inject Git commmit details through the docker build so that the we can output that in our metrics, logs, etc. I am not 100% sure this was ever configured correctly, but I wanted to call out that this would seem to be a gap. Otherwise, this looks good to me.

The goreleaser path injects main.GitCommit, main.GitDate, main.Version via ldflags. melange/proxyd.yaml:29 uses only -s -w and package.version: "0.0.0".

PR #665, opened 2026-07-07
proxyd: parallelize EL consensus-poller RPCs in UpdateBackend

danielllek:
**Description**

UpdateBackend previously issued eth_syncing, eth_getBlockByNumber("latest"), ("safe"), and ("finalized") sequentially against the same backend, so one poll cycle cost the sum of four round trips. In our production deployment this measured at ~690ms/poll even with consensus_poller_interval set to 250ms, since the call chain itself exceeded the configured interval and became the real bottleneck.

Fire the four RPCs concurrently instead; a cycle now costs roughly one round trip. Error/zero-value precedence on failure is unchanged from the sequential version (sync error > latest > safe > finalized), just evaluated after all four results are in rather than short-circuiting between calls.

Also fixes MockBackend.wrappedHandler in the integration test harness, which held its mutex across the full handler invocation and serialized concurrent requests to the same mock backend — harmless for the old sequential poller, but it caused the new concurrent calls to queue behind each other and time out in tests that simulate a slow backend. The lock now only guards the shared request log and handler reference, not handler execution, matching how a real backend actually behaves under concurrent load.

**Tests**

- `go test ./...` (full `proxyd` module) — pass
- `go test ./... -race` — pass, no data races introduced by the new concurrent goroutines
- `go test ./integration_tests/... -run TestConsensus -count=3` — repeated to rule out flakiness from the new concurrency, pass
- Added a new case ("prevent using a backend when eth_syncing errors") covering an actual RPC/transport-level failure on the concurrent isELInSync path, distinct from the existing not-in-sync case which only covers a successful-but-not-synced response
- Manual end-to-end smoke test: built the Docker image from this branch and ran it with a config pointed at real production backends. Confirmed it forms a healthy 3-backend consensus group, serves `eth_blockNumber` correctly, shows no bans/errors, and reports a lower `proxyd_consensus_backend_update_delay` than the previous sequential implementation under the same `consensus_poller_interval`.

The existing `TestConsensus` suite in `integration_tests/` already exercised `UpdateBackend`'s success/error/zero-block paths end-to-end through the mock backends and passes unchanged (the mock harness fix above was needed for it to keep doing so under real concurrency instead of only proving it works when serialized).

**Additional context**

Root-caused via production instrumentation rather than guesswork: `proxyd_consensus_backend_update_delay` is `time.Since(lastUpdate)` recorded on every poll (see `RecordConsensusBackendUpdateDelay` in `metrics.go`), so it directly reports the achieved loop period, not a random-phase staleness sample. In our deployment, it measured ~690-694ms with `consensus_poller_interval = "250ms"` configured — meaning the four sequential RPCs inside `UpdateBackend`, not the poller's idle wait, were the actual bottleneck. This change removes that bottleneck; it does not change proxyd's consensus semantics (the served head is still `min(latest_block)` across in-consensus backends).

**Metadata**

Fixes #[Link to Issue]

### DIFF
--- a/proxyd/consensus_poller.go
+++ b/proxyd/consensus_poller.go
@@ -3,7 +3,6 @@ package proxyd
 import (
 	"context"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"strconv"
 	"strings"
@@ -16,12 +15,6 @@ import (
 
 const (
 	DefaultPollerInterval = 1 * time.Second
-)
-
-var (
-	errZeroLatestBlock    = errors.New("backend responded with blockheight 0 for latest block")
-	errZeroSafeBlock      = errors.New("backend responded with blockheight 0 for safe block")
-	errZeroFinalizedBlock = errors.New("backend responded with blockheight 0 for finalized block")
 )
 
 type OnConsensusBroken func()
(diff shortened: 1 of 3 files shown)


danielllek:
No real gain after deep tests :-/

PR #676, opened 2026-07-27
proxyd: add per-RPC-call metrics for usage and error-rate SLO

zhwrd:
## What

Adds two proxyd metrics that count **individual JSON-RPC calls** with the outcome the client actually received:

- `proxyd_rpc_calls_total{backend_name, method_name, status_code, transport}` — one increment per client-initiated call, recorded after retries and failover.
- `proxyd_rpc_notifications_total{backend_name}` — client-bound WS subscription notifications.

Also fixes two pre-existing bugs found while instrumenting: `proxyd_http_response_codes_total` counted nothing for successful batches (Change, item 4), and batch-size rejections were answered HTTP 200 (Change, item 5).

Nothing consumes the new metrics yet — no dashboard or SLO is repointed in this PR. See Rollout below.

> [!IMPORTANT]
> **Client-visible status code change: batches over `max_batch_size` now return HTTP 413 instead of HTTP 200.**
>
> `ErrTooManyBatchRequests` carried no `HTTPErrorCode`, so proxyd answered a rejected batch with HTTP **200** and a JSON-RPC error body. It is now 413, matching `ErrRequestBodyTooLarge`.
>
> **Who this affects:** any client sending batches larger than the deployment's `max_batch_size`. The JSON-RPC error body is unchanged — only the HTTP status. Clients that key off the JSON-RPC error are unaffected. Clients that treat non-200 as retryable will begin retrying a rejection that can never succeed, and clients that only inspect the body on a 2xx will now surface a transport-level error instead.
>
> **Why it has to change:** as HTTP 200 this rejection was invisible to every error-rate metric. The rejection also returns before `handleBatchRPC`, so the new per-call metric cannot compensate for it either — a client whose every batch was rejected failed 100% of the time while both metrics reported health. It was the only pre-dispatch rejection landing in the 2xx bucket (siblings: parse error 400, empty batch 400, body too large 413, drain 503).
>
> **Needs a release note.** `integration_tests/batching_test.go` asserted the old 200 — which is how we know nothing internal depends on it — and now asserts 413.

## Why

The OP Enterprise dashboard measures RPC usage and RPC error rate from metrics that cannot express either quantity, because proxyd has no metric representing "an RPC call the customer made, and the outcome they received." Every existing metric is either per-HTTP-envelope or per-backend-attempt:

- **`proxyd_rpc_requests_total` undercounts batch traffic by up to 100x.** It increments once per *minibatch forwarded upstream* (`backend.go`, `BackendGroup.ForwardWithSource`), so with `max_upstream_batch_size = 100` a 100-call batch counted as 1. It also misses cache hits entirely, counts batches containing zero real RPCs, and multi-counts batches spanning backend groups. This is the metric monthly quota is measured against.
- **`proxyd_http_response_codes_total` counted zero for every successful batch.** `writeBatchRPCRes` omitted the increment that `writeRPCRes` performs. Batches that failed wholesale *were* counted, so batch traffic entered the error-rate numerator but never the denominator — biasing the measured rate upward.
- **`proxyd_rpc_backend_http_response_codes_total` counts backend round-trips, not customer outcomes.** It includes consensus-poller traffic, double-counts retries, records failed-over attempts the customer never saw, and misses transport errors entirely (`doForward` returns before the increment on `err != nil`).
- **WebSocket traffic contributed to none of the status-code, error-rate, or latency metrics**, despite WS being enabled on every customer-facing deployment.

A batch also always returns HTTP 200 at the envelope level even when every call inside it failed, so an envelope-level SLI cannot fail for batch traffic at all.

## Change

1. **Status derivation** (`metrics.go`). `statusCodeForRPCRes` maps one `*RPCRes` to an HTTP-equivalent status. The load-bearing rule: an error with `HTTPErrorCode == 0` maps to **400, never 500**. That is the case where a backend answered HTTP 200 carrying a JSON-RPC error (`already known`, `nonce too low`, `execution reverted`) — `doForward` only stamps `HTTPErrorCode` on a non-200 upstream status. Mapping those to 5xx would burn the error-rate SLO on routine client faults.

2. **HTTP instrumentation** (`server.go`, `handleBatchRPC`). Per-element attribution arrays parallel to `responses`, recorded once at each client-visible exit. This one site covers batch *and* single requests (`HandleRPC` calls it with a one-element slice), and is the only place with per-element indices, the serving backend, post-failover outcomes, cache hits, and proxyd-local errors all in scope. Sentinel `backend_name` values: `cache` for cache hits, `proxyd` for locally-answered or rejected calls.

3. **WS instrumentation** (`server.go`, `backend.go`). `handleWSRPC` split into a recording wrapper plus an inner function returning `servedBy`, so its five return paths record exactly once. `clientPump` records the relay path (`eth_subscribe`/`eth_unsubscribe`) at send time with `status_code="relayed"`, since those responses return asynchronously in `backendPump` with no id correlation. `backendPump` classifies backend messages structurally — `method` with no `id` is a notification; a message carrying an `id` is already counted and is not double-counted.

4. **Envelope-metric fix** (`server.go`). `writeBatchRPCRes` now increments `proxyd_http_response_codes_total` once per response (not per element), and the `HandleRPC` drain path is counted. Healthz stays uninstrumented so probe traffic does not dilute the denominator.

5. **`ErrTooManyBatchRequests` gains `HTTPErrorCode: 413`** (`backend.go`). One line, but it is the client-visible change called out at the top of this description. See that callout for who it affects and why it cannot stay at 200.

Two properties worth calling out:

- **Attribution is per element, including for locally-overridden calls.** When consensus or block-tag rewriting answers some elements of a batch and a backend answers the rest, the overridden elements must not inherit the serving backend's name. `OverrideResponses` — the single point where overrides merge back into a batch result — marks each one with an unexported `servedLocally` flag, and `handleBatchRPC` attributes those to `proxyd`. The flag is absent from `rpcResJSON`, so it never reaches the wire.
- **The consensus poller is excluded structurally, not by label filtering.** It reaches backends via `ForwardRPC` → `doForward`, entering none of the three instrumentation sites. This is the same reason the existing latency summary is already poller-clean, and it means no `origin` label is needed on the new metric.
- **`method_name` cardinality is bounded at every site that sets it.** Method names are client-supplied and reachable by unauthenticated clients, so anything outside the configured whitelist collapses to `method_not_allowed`. `eth_accounts` is passed through explicitly because it is answered locally before the mappings lookup and would otherwise be recorded as a *successful* `method_not_allowed`.

`status_code="relayed"` is deliberately non-numeric so `status_code=~"2.."` and `=~"5.."` can never match a call whose outcome is unknowable, and it is excluded from the SLO denominator.

No signature of `Forward`, `ForwardWithSource`, `ForwardRPC`, `doForward`, or any `BackendGroup` method changed — this deliberately avoids a breaking *Go API* change for external callers. The one client-visible behavior change is the 413 above.

## Testing

Unit tests cover status derivation against the real error table (including the 200-with-JSON-RPC-error case asserting 400, not 500), the cardinality bounds, the batch-vs-single deadline status divergence, and the `servedLocally` marking at the override merge point. Integration tests cover: a 3-element batch producing exactly 3 increments (previously 0); a 12-element mixed batch spanning two minibatches, asserting all 12 increments plus per-status and per-backend attribution; cache-hit attribution to `cache`; consensus-poller traffic producing zero increments (with a guard asserting the poller actually reached the backend, so the zero cannot hold vacuously); WS relay + notification counting; the WS mapped-method path; and that a client naming its method `unknown` still collapses to `method_not_allowed`.

Two gaps worth naming rather than hiding. The single-request deadline short-circuit is covered at the helper, not end to end — reaching it requires the deadline to expire inside `prepareRPCForForward`, since a one-element request has only one minibatch iteration. And the partial-override attribution branch in `handleBatchRPC` is covered at the merge point, not through a `consensus_aware` group with a block tag past the consensus latest.

Two behaviors are mutation-verified — the instrumentation was temporarily removed and the tests confirmed to fail (`expected: 3, actual: 0` for the batch count; the cardinality bound likewise) before reverting.

Full module suite passes uncached: unit ~22s, `integration_tests` ~99s. `gofmt` and `go vet` clean.

## Rollout

**This PR changes no dashboard or SLO**, but it does change one client-visible status code. In order:

0. **Release-note the 413 before deploy.** Batches over `max_batch_size` stop returning HTTP 200 (see the callout at the top). This is the only item that affects clients rather than dashboards, and it ships the moment this deploys — it cannot be staged behind the bake window like the rest. Worth checking whether any known integration sends oversized batches; if one does, it will start seeing a transport-level error where it previously saw a 200 with an error body.
1. **Bake alongside the existing metrics** and compare per chain: new vs `proxyd_rpc_requests_total`, and the new 5xx ratio vs the current SLO value.
2. **Wire the per-call SLI alert before the bake window opens, not after.** See item 3 — during the bake there is a window in which batch failures are *less* visible on the existing metric than they were before this PR, and the new metric has no alert on it yet. That window should be as short as possible.
3. **`proxyd_http_response_codes_total` semantics change in this PR, and not only in the benign direction.** It now counts successful batches where it previously counted nothing, so the denominator grows and any existing error-rate expression over it will see its measured value **drop** on deploy. That much is the intended correction. The sharper edge: a batch is always HTTP 200 at the envelope level, so an outage that leaves batches *structurally* intact — every element `ErrNoBackends` — now adds only to `{status_code="200"}`. Previously those batches reached the numerator via `writeRPCError` and nothing reached the denominator; now the denominator grows while element-level failures still contribute nothing, so **the measured 5xx ratio over this metric falls during exactly that outage**. `proxyd_rpc_calls_total` is the metric that catches it. Both effects warrant a Grafana annotation on deploy day, and it will read as a step change to whoever owns the alerts in the `k8s` repo.
4. **Commercial review of the usage delta before the usage tile repoints.** Correcting an up-to-100x batch undercount changes the number customers are measured against. This is a pricing decision, not an engineering one.

Two things to carry into the dashboard/SLO work:

- The provider-filter allowlist in the dashboard must gain `proxyd` and `cache`, or the worst outage class (`ErrNoBackends`, rate limits, rejections — all attributed to `proxyd`) and all cache hits will be invisible in the provider-filtered graph.
- The new per-call SLI is structurally blind to envelope-level rejections (drain, 401, 413, batch-parse error), because those fail before any individual call is parsed and dispatched. If proxyd rejects everything, the numerator *and* denominator are empty, so the SLI reads 0% error or NaN rather than 100%. **Keep** an availability alert on `proxyd_http_response_codes_total` alongside the new SLI rather than replacing it.

## Relationship to #673

[#673](https://github.com/ethereum-optimism/infra/pull/673) adds an `origin` label to `proxyd_rpc_backend_http_response_codes_total` so consensus-poller traffic can be filtered out of the existing panel. It is orthogonal and can land independently. This PR subsumes the underlying need: the new metric excludes the poller structurally, and also addresses batching, retry double-counting, failed-over requests, and WS, none of which the `origin` label reaches.

### DIFF
--- a/proxyd/backend.go
+++ b/proxyd/backend.go
@@ -1479,6 +1479,18 @@ func (w *WSProxier) Proxy(ctx context.Context) error {
 	return err
 }
 
+// metricMethodName bounds the method_name label cardinality for WS traffic. See
+// Server.metricMethodName — WS uses its own whitelist.
+func (w *WSProxier) metricMethodName(method string) string {
+	if method == "" {
+		return MethodUnknown
+	}
+	if !w.methodWhitelist.Has(method) {
+		return MethodNotAllowed
+	}
+	return method
+}
+
 func (w *WSProxier) clientPump(ctx context.Context, errC chan error) {
 	for {
 		// Block until we get a message.
@@ -1526,6 +1538,7 @@ func (w *WSProxier) clientPump(ctx context.Context, errC chan error) {
 			)
 			msg = mustMarshalJSON(NewRPCErrorRes(id, err))
 			RecordRPCError(ctx, BackendProxyd, method, err)
+			RecordRPCCall(BackendProxyd, w.metricMethodName(method), statusCodeForRPCRes(NewRPCErrorRes(id, err)), RPCRequestSourceWS)
 
 			// Send error response to client
 			err = w.writeClientConn(msgType, msg)
@@ -1540,6 +1553,7 @@ func (w *WSProxier) clientPump(ctx context.Context, errC chan error) {
 		if req.Method == "eth_accounts" {
 			msg = mustMarshalJSON(NewRPCRes(req.ID, emptyArrayResponse))
 			RecordRPCForward(ctx, BackendProxyd, "eth_accounts", RPCRequestSourceWS)
+			RecordRPCCall(BackendProxyd, "eth_accounts", statusCodeOK, RPCRequestSourceWS)
 			err = w.writeClientConn(msgType, msg)
 			if err != nil {
 				errC <- err
@@ -1549,9 +1563,11 @@ func (w *WSProxier) clientPump(ctx context.Context, errC chan error) {
 		}
 
 		if w.requestHandler != nil && w.requestHandler(req) {
+			// handleWSRPC records this call; do not record it here too.
 			if !w.acquireRequestSlot() {
 				msg = mustMarshalJSON(NewRPCErrorRes(req.ID, ErrTooManyRequests))
 				RecordRPCError(ctx, BackendProxyd, req.Method, ErrTooManyRequests)
+				RecordRPCCall(BackendProxyd, w.metricMethodName(req.Method), statusCodeForRPCRes(NewRPCErrorRes(req.ID, ErrTooManyRequests)), RPCRequestSourceWS)
 				err = w.writeClientConn(msgType, msg)
 				if err != nil {
 					errC <- err
@@ -1564,6 +1580,11 @@ func (w *WSProxier) clientPump(ctx context.Context, errC chan error) {
 		}
 
 		RecordRPCForward(ctx, w.backend.Name, req.Method, RPCRequestSourceWS)
+		// Counted at send time: this call's response returns asynchronously in
+		// backendPump with no id correlation back to this request, so its outcome
+		// is unknowable here. Counting at send time guarantees usage completeness
+		// at the cost of an unknown status.
+		RecordRPCCall(w.backend.Name, w.metricMethodName(req.Method), StatusCodeRelayed, RPCRequestSourceWS)
 		log.Info(
 			"forwarded WS message to backend",
 			"method", req.Method,
@@ -1654,6 +1675,13 @@ func (w *WSProxier) backendPump(ctx context.Context, errC chan error) {
 			msg = mustMarshalJSON(NewRPCErrorRes(id, err))
 			log.Info("backend responded with error", "err", err)
 		} else {
+			// A backend message with no id is a subscription notification pushed by
+			// the backend, not a response to a client call. A message WITH an id is
+			// a response to a relayed call, already counted at send time in
+			// clientPump — counting it here would double-count.
+			if len(res.ID) == 0 {
+				RecordRPCNotification(w.backend.Name)
+			}
 			if res.IsError() {
 				log.Info(
 					"backend responded with RPC error",
--- a/proxyd/backend.go
+++ b/proxyd/backend.go
@@ -1485,6 +1485,16 @@ func (w *WSProxier) metricMethodName(method string) string {
 	if method == "" {
 		return MethodUnknown
 	}
+	// MethodUnknown and MethodNotAllowed are already-bounded sentinels — e.g.
+	// clientPump passes MethodUnknown through here when a frame fails
+	// ParseRPCReq. Pass them through unchanged rather than re-running them
+	// through the whitelist, where the literal strings "unknown" and
+	// "method_not_allowed" would themselves fail the Has check and get
+	// relabeled MethodNotAllowed, mislabeling a parse failure as a
+	// not-allowed method.
+	if method == MethodUnknown || method == MethodNotAllowed {
+		return method
+	}
 	if !w.methodWhitelist.Has(method) {
 		return MethodNotAllowed
 	}
@@ -1563,7 +1573,6 @@ func (w *WSProxier) clientPump(ctx context.Context, errC chan error) {
 		}
 
 		if w.requestHandler != nil && w.requestHandler(req) {
-			// handleWSRPC records this call; do not record it here too.
 			if !w.acquireRequestSlot() {
 				msg = mustMarshalJSON(NewRPCErrorRes(req.ID, ErrTooManyRequests))
 				RecordRPCError(ctx, BackendProxyd, req.Method, ErrTooManyRequests)
@@ -1575,6 +1584,7 @@ func (w *WSProxier) clientPump(ctx context.Context, errC chan error) {
 				}
 				continue
 			}
+			// handleWSRPC records this call; do not record it here too.
 			go w.processClientRPC(ctx, msgType, req, errC)
 			continue
 		}
--- a/proxyd/backend.go
+++ b/proxyd/backend.go
@@ -1485,16 +1485,6 @@ func (w *WSProxier) metricMethodName(method string) string {
 	if method == "" {
 		return MethodUnknown
 	}
-	// MethodUnknown and MethodNotAllowed are already-bounded sentinels — e.g.
-	// clientPump passes MethodUnknown through here when a frame fails
-	// ParseRPCReq. Pass them through unchanged rather than re-running them
-	// through the whitelist, where the literal strings "unknown" and
-	// "method_not_allowed" would themselves fail the Has check and get
-	// relabeled MethodNotAllowed, mislabeling a parse failure as a
-	// not-allowed method.
-	if method == MethodUnknown || method == MethodNotAllowed {
-		return method
-	}
 	if !w.methodWhitelist.Has(method) {
 		return MethodNotAllowed
 	}
@@ -1535,10 +1525,22 @@ func (w *WSProxier) clientPump(ctx context.Context, errC chan error) {
 		req, err := w.prepareClientMsg(msg)
 		if err != nil {
 			var id json.RawMessage
+			// method is the raw client-supplied string, used only for
+			// RecordRPCError below, which deliberately accepts unbounded
+			// cardinality. It must NOT be passed to metricMethodName: an
+			// arbitrary client-supplied method (e.g. "unknown") must not be
+			// able to masquerade as the MethodUnknown sentinel and evade the
+			// method_not_allowed bucket. metricLabel below is derived from
+			// req instead, since prepareClientMsg has exactly two failure
+			// modes: req == nil means ParseRPCReq failed (no method to speak
+			// of); req != nil means the method parsed but failed the
+			// whitelist check.
 			method := MethodUnknown
+			metricLabel := MethodUnknown
 			if req != nil {
 				id = req.ID
 				method = req.Method
+				metricLabel = MethodNotAllowed
 			}
 			log.Info(
 				"error preparing client message",
@@ -1548,7 +1550,7 @@ func (w *WSProxier) clientPump(ctx context.Context, errC chan error) {
 			)
 			msg = mustMarshalJSON(NewRPCErrorRes(id, err))
 			RecordRPCError(ctx, BackendProxyd, method, err)
-			RecordRPCCall(BackendProxyd, w.metricMethodName(method), statusCodeForRPCRes(NewRPCErrorRes(id, err)), RPCRequestSourceWS)
+			RecordRPCCall(BackendProxyd, metricLabel, statusCodeForRPCRes(NewRPCErrorRes(id, err)), RPCRequestSourceWS)
 
 			// Send error response to client
 			err = w.writeClientConn(msgType, msg)
--- a/proxyd/backend.go
+++ b/proxyd/backend.go
@@ -79,6 +79,13 @@ var (
 	ErrTooManyBatchRequests = &RPCErr{
 		Code:    JSONRPCErrorInternal - 14,
 		Message: "too many RPC calls in batch request",
+		// 413 matches ErrRequestBodyTooLarge, the sibling "request is too big"
+		// rejection, and is a client-visible change: this was previously HTTP 200.
+		// Without an HTTPErrorCode this rejection was written as HTTP 200, so a
+		// client whose every batch exceeds max_batch_size failed 100% of the time
+		// while both proxyd_http_response_codes_total and (because the rejection
+		// returns before handleBatchRPC) proxyd_rpc_calls_total reported health.
+		HTTPErrorCode: 413,
 	}
 	ErrGatewayTimeout = &RPCErr{
 		Code:          JSONRPCErrorInternal - 15,
@@ -1479,8 +1486,26 @@ func (w *WSProxier) Proxy(ctx context.Context) error {
 	return err
 }
 
-// metricMethodName bounds the method_name label cardinality for WS traffic. See
-// Server.metricMethodName — WS uses its own whitelist.
+// metricMethodName bounds the method_name label cardinality for WS traffic by
+// collapsing anything outside ws_method_whitelist to a sentinel.
+//
+// Both current call sites in clientPump are downstream of a successful
+// prepareClientMsg, which already applies the same whitelist, so the collapsing
+// branches are unreachable today and only the tests exercise them. This is kept
+// deliberately: the bound is self-contained (it consults w.methodWhitelist
+// directly, not the caller's gate), method names are client-supplied and
+// reachable by unauthenticated clients, and unbounded Prometheus label
+// cardinality is the failure mode. Any future WS metric site can call this
+// without having to prove where it sits relative to the whitelist gate.
+//
+// Note this whitelist is not the one Server.metricMethodName uses: that one keys
+// on rpc_method_mappings. shouldHandleWSRPC admits a request via
+// txFilter.IsSubmissionMethod or txValidationMethods without consulting
+// rpc_method_mappings, so a method in ws_method_whitelist but absent from
+// rpc_method_mappings records under its real name when rejected by the semaphore
+// here and as method_not_allowed when handleWSRPCInner rejects it. Both are
+// rejections, so no call is miscounted, but sum by (method_name) is not
+// comparable across the two paths in that (unusual) configuration.
 func (w *WSProxier) metricMethodName(method string) string {
 	if method == "" {
 		return MethodUnknown
@@ -1548,9 +1573,12 @@ func (w *WSProxier) clientPump(ctx context.Context, errC chan error) {
 				"req_id", GetReqID(ctx),
 				"err", err,
 			)
-			msg = mustMarshalJSON(NewRPCErrorRes(id, err))
+			// Build the response once and derive the metric from that same value, so
+			// the recorded status cannot drift from what the client receives.
+			errRes := NewRPCErrorRes(id, err)
+			msg = mustMarshalJSON(errRes)
 			RecordRPCError(ctx, BackendProxyd, method, err)
-			RecordRPCCall(BackendProxyd, metricLabel, statusCodeForRPCRes(NewRPCErrorRes(id, err)), RPCRequestSourceWS)
+			RecordRPCCall(BackendProxyd, metricLabel, statusCodeForRPCRes(errRes), RPCRequestSourceWS)
 
 			// Send error response to client
 			err = w.writeClientConn(msgType, msg)
@@ -1576,9 +1604,10 @@ func (w *WSProxier) clientPump(ctx context.Context, errC chan error) {
 
 		if w.requestHandler != nil && w.requestHandler(req) {
 			if !w.acquireRequestSlot() {
-				msg = mustMarshalJSON(NewRPCErrorRes(req.ID, ErrTooManyRequests))
+				errRes := NewRPCErrorRes(req.ID, ErrTooManyRequests)
+				msg = mustMarshalJSON(errRes)
 				RecordRPCError(ctx, BackendProxyd, req.Method, ErrTooManyRequests)
-				RecordRPCCall(BackendProxyd, w.metricMethodName(req.Method), statusCodeForRPCRes(NewRPCErrorRes(req.ID, ErrTooManyRequests)), RPCRequestSourceWS)
+				RecordRPCCall(BackendProxyd, w.metricMethodName(req.Method), statusCodeForRPCRes(errRes), RPCRequestSourceWS)
 				err = w.writeClientConn(msgType, msg)
 				if err != nil {
 					errC <- err
@@ -1592,11 +1621,6 @@ func (w *WSProxier) clientPump(ctx context.Context, errC chan error) {
 		}
 
 		RecordRPCForward(ctx, w.backend.Name, req.Method, RPCRequestSourceWS)
-		// Counted at send time: this call's response returns asynchronously in
-		// backendPump with no id correlation back to this request, so its outcome
-		// is unknowable here. Counting at send time guarantees usage completeness
-		// at the cost of an unknown status.
-		RecordRPCCall(w.backend.Name, w.metricMethodName(req.Method), StatusCodeRelayed, RPCRequestSourceWS)
 		log.Info(
 			"forwarded WS message to backend",
 			"method", req.Method,
@@ -1609,6 +1633,14 @@ func (w *WSProxier) clientPump(ctx context.Context, errC chan error) {
 			errC <- err
 			return
 		}
+
+		// Recorded after the write succeeds, so "relayed" means actually relayed —
+		// a call lost to a failed backend write is not counted as one proxyd
+		// delivered. Counted here rather than on the response because this call's
+		// response returns asynchronously in backendPump with no id correlation back
+		// to this request, so its outcome is unknowable on this side: counting at
+		// send time guarantees usage completeness at the cost of an unknown status.
+		RecordRPCCall(w.backend.Name, w.metricMethodName(req.Method), StatusCodeRelayed, RPCRequestSourceWS)
 	}
 }
 
@@ -1679,6 +1711,7 @@ func (w *WSProxier) backendPump(ctx context.Context, errC chan error) {
 		}
 
 		res, err := w.parseBackendMsg(msg)
+		isNotification := false
 		if err != nil {
 			var id json.RawMessage
 			if res != nil {
@@ -1687,13 +1720,13 @@ func (w *WSProxier) backendPump(ctx context.Context, errC chan error) {
 			msg = mustMarshalJSON(NewRPCErrorRes(id, err))
 			log.Info("backend responded with error", "err", err)
 		} else {
-			// A backend message with no id is a subscription notification pushed by
-			// the backend, not a response to a client call. A message WITH an id is
-			// a response to a relayed call, already counted at send time in
-			// clientPump — counting it here would double-count.
-			if len(res.ID) == 0 {
-				RecordRPCNotification(w.backend.Name)
-			}
+			// A backend message with no id and no error is a subscription notification
+			// pushed by the backend, not a response to a client call. A message WITH an
+			// id is a response to a relayed call, already counted at send time in
+			// clientPump — counting it here would double-count. parseBackendMsg does no
+			// shape validation beyond JSON parsing, so an id-less *error* frame is also
+			// possible and is not a notification either.
+			isNotification = len(res.ID) == 0 && !res.IsError()
 			if res.IsError() {
 				log.Info(
 					"backend responded with RPC error",
@@ -1717,6 +1750,13 @@ func (w *WSProxier) backendPump(ctx context.Context, errC chan error) {
 		if err != nil {
 			errC <- err
 			return
+		}
+
+		// Recorded only after the client write succeeds: the metric claims delivery,
+		// so a notification dropped on a slow or disconnected client must not read as
+		// one the subscriber received.
+		if isNotification {
+			RecordRPCNotification(w.backend.Name)
 		}
 	}
 }
@@ -1983,6 +2023,10 @@ func (bg *BackendGroup) ForwardRequestToBackendGroup(
 
 func OverrideResponses(res []*RPCRes, overriddenResponses []*indexedReqRes) []*RPCRes {
 	for _, ov := range overriddenResponses {
+		// proxyd produced this response itself; no backend saw the request. Mark it
+		// so per-call metrics do not attribute it to whichever backend served the
+		// elements that were forwarded alongside it.
+		ov.res.servedLocally = true
 		if len(res) > 0 {
 			// insert ov.res at position ov.index
 			res = append(res[:ov.index], append([]*RPCRes{ov.res}, res[ov.index:]...)...)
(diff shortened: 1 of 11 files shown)


proxyd/server.go
```diff
@@ -732,6 +767,16 @@ func (s *Server) handleBatchRPC(ctx context.Context, reqs []json.RawMessage, isL
...
+				// The whole request set failed wholesale: this function returns a bare
+				// error, so HandleRPC writes a single error envelope for the entire
+				// batch (ErrGatewayTimeout/504 on the batch path; the single-request
+				// path has no DeadlineExceeded case and writes ErrInternal/500 instead).
+				// The client receives none of the per-element results computed so far
+				// (cache hits, locally-answered elements, or minibatches already
+				// forwarded in a prior iteration) — every element must be recorded with
+				// the fallback status, not the per-element status that never reached
+				// the client.
+				recordRPCCallsAll(callMethods, callBackends, RPCRequestSourceHTTP, statusCodeGatewayTimeout)
```
zhwrd: **Deadline short-circuit records 504, but the single-request path writes 500 to the client.**

`recordRPCCallsAll(..., statusCodeGatewayTimeout)` is unconditional, yet only the *batch* caller maps `context.DeadlineExceeded` to `ErrGatewayTimeout` (server.go:448-451). The non-batch caller (server.go:470-479) has no `DeadlineExceeded` branch and falls through to `writeRPCError(ctx, w, nil, ErrInternal)` → HTTP 500.

The comment on lines 771-773 names the divergence but the code doesn't act on it — and this breaks `recordRPCCalls`' own documented contract that `fallbackStatus` "must match the status the HTTP handler goes on to write to the client."

Repro: a single (non-batch) `eth_sendRawTransaction` whose `txFilter.Apply` interop round-trip burns `s.timeout`. Client gets 500, `http_response_codes_total` records 500, `rpc_calls_total` records **504**. A 504-specific timeout alert fires for what operators see as an internal error, or a 500-only alert misses it entirely.

`handleBatchRPC` already receives `isBatch`:

```suggestion
				fallbackStatus := statusCodeInternal
				if isBatch {
					// HandleRPC maps context.DeadlineExceeded to ErrGatewayTimeout on the
					// batch path only; the single-request path writes ErrInternal.
					fallbackStatus = statusCodeGatewayTimeout
				}
				recordRPCCallsAll(callMethods, callBackends, RPCRequestSourceHTTP, fallbackStatus)
```

Worth an integration test on this path — `metrics_test.go` only exercises the helper directly, so the mismatch can't be caught today.

proxyd/server.go
```diff
@@ -776,8 +826,12 @@ func (s *Server) handleBatchRPC(ctx context.Context, reqs []json.RawMessage, isL
 				}
 			}
 
+			// sb is "" when the forward failed outright, which backendNameFromServedBy
+			// resolves to BackendProxyd — such a batch was never served by a backend.
+			forwardedBy := backendNameFromServedBy(sb)
```
zhwrd, line 835: **The comment is factually wrong: `sb == ""` also happens on fully-successful responses, so those get attributed to `proxyd`.**

`ForwardWithSource` returns `("", nil)` with a populated `res` when consensus/block-tag rewriting overrode *every* element (backend.go:1131-1133):

```go
if len(rpcReqs) == 0 {
    return OverrideResponses(nil, overriddenResponses), "", nil
}
```

That's reachable in production: `optimism_syncStatus` on a `consensus_aware` group in CL mode is served entirely from `bg.Consensus.GetConsensusSyncStatusBody()` (backend.go:2009-2028) and consumes every request. Same for a batch where all elements hit `RewriteOverride`/`RewriteOverrideError`. Those calls record `backend_name="proxyd"`, `status_code="200"` — consensus-served traffic silently vanishes from per-backend call counts and inflates the "proxyd answered locally" bucket.

The mirror case is worse: with a *partial* override (e.g. `[eth_getBlockByNumber(out-of-range), eth_call]`), `sb` is the real backend and the loop below stamps `forwardedBy` on **both** indices — so an element proxyd answered locally is attributed to a backend that never saw it.

`ForwardWithSource` can't distinguish these from the caller today. Either return the override count alongside `servedBy`, or attribute per-element by checking whether the response came from `overriddenResponses`.

proxyd/server.go
```diff
@@ -1168,6 +1242,7 @@ func writeBatchRPCRes(ctx context.Context, w http.ResponseWriter, res []*RPCRes)
 		RecordRPCError(ctx, BackendProxyd, MethodUnknown, err)
 		return
 	}
+	httpResponseCodesTotal.WithLabelValues(strconv.Itoa(200)).Inc()
```
zhwrd: **Hardcoded 200 means batch traffic can only ever grow the denominator, never the numerator.**

`writeBatchRPCRes` hardcodes `WriteHeader(200)` and now hardcodes the matching increment, regardless of per-element outcomes. Combined with the fix, a backend outage that leaves batches *structurally* intact (every element `ErrNoBackends`) adds only to `{status_code="200"}` — so the measured 5xx ratio over this metric **falls** during the outage.

The PR body frames this as "the denominator grows, measured value drops on deploy," which is right, but the outage case is the sharper edge: previously batch failures reached the numerator via `writeRPCError` (504 at server.go:449, 500 at server.go:459) and nothing reached the denominator; now the denominator grows while element-level failures still contribute nothing. During the bake window — where "nothing consumes the new metrics yet" — there is no per-call SLI wired up to compensate.

Suggest calling this out explicitly in the Grafana annotation, and prioritizing the per-call SLI alert so the gap doesn't sit open. (Also: `statusCodeOK` is defined right there in metrics.go — no need for `strconv.Itoa(200)`.)

proxyd/metrics.go
```diff
@@ -171,6 +183,28 @@ var (
 		"status_code",
 	})
 
+	rpcCallsTotal = promauto.NewCounterVec(prometheus.CounterOpts{
+		Namespace: MetricsNamespace,
+		Name:      "rpc_calls_total",
+		Help: "Count of client-initiated JSON-RPC calls by the outcome the client received. " +
+			"One increment per call: batch elements are counted individually, as is each WS frame. " +
+			"Recorded after retries and failover, so it reflects what the client saw. " +
+			"Excludes consensus-poller traffic.",
```
zhwrd, line 192: **Denominator gap: `ErrTooManyBatchRequests` returns HTTP 200 *and* zero `rpc_calls_total`, so both metrics read healthy during a 100%-failure condition.**

The PR body's mitigation is "keep an availability alert on `proxyd_http_response_codes_total` alongside the new SLI." That backstop has a hole. `ErrTooManyBatchRequests` (backend.go:79-82) is declared with no `HTTPErrorCode`:

```go
ErrTooManyBatchRequests = &RPCErr{
    Code:    JSONRPCErrorInternal - 14,
    Message: "too many RPC calls in batch request",
}
```

so `writeRPCRes` leaves `statusCode := 200` (server.go:1217-1219) and increments `http_response_codes_total{status_code="200"}`. The rejection at server.go:436-439 also returns before `handleBatchRPC`, so `rpc_calls_total` gets nothing.

Net: a client whose every batch exceeds `max_batch_size` fails 100% of the time while **both** metrics report health. The sibling pre-dispatch rejections are all fine (parse error → 400, empty batch → 400, body too large → 413, drain → 503 as of this PR) — this is the only one that lands in the 2xx bucket.

Pre-existing for the HTTP metric, but newly load-bearing given this PR's rollout plan. Cheapest fix is a one-line `HTTPErrorCode: 413` (or 400) on `ErrTooManyBatchRequests`.

proxyd/backend.go
```diff
@@ -1654,6 +1687,13 @@ func (w *WSProxier) backendPump(ctx context.Context, errC chan error) {
 			msg = mustMarshalJSON(NewRPCErrorRes(id, err))
 			log.Info("backend responded with error", "err", err)
 		} else {
+			// A backend message with no id is a subscription notification pushed by
+			// the backend, not a response to a client call. A message WITH an id is
+			// a response to a relayed call, already counted at send time in
+			// clientPump — counting it here would double-count.
+			if len(res.ID) == 0 {
```
zhwrd: **`len(res.ID) == 0` is a purely structural test — it classifies things that aren't notifications, and counts before delivery.**

`ParseRPCRes` (rpc.go:105-117) is just `io.ReadAll` + `json.Unmarshal` into `RPCRes` with no validation. It checks neither the `jsonrpc` version nor the `method`/`params` shape that actually makes a frame a notification, so **any** parseable object lacking an `id` key increments the counter.

Two concrete consequences:

1. **Client-drivable.** `prepareClientMsg` (backend.go:1738-1748) only does `ParseRPCReq` + whitelist — it never calls `ValidateRPCReq`, which is the thing that rejects a missing id. So a client can relay `{"jsonrpc":"2.0","method":"eth_subscribe","params":["newHeads"]}` with no `id`; the backend's id-less reply is then booked as a subscription notification. An unauthenticated client can inflate `rpc_notifications_total` on demand.
2. **"Delivered" is wrong.** The increment is here, but `w.writeClientConn` is at line 1716. A slow/disconnected client whose write fails still gets counted, so a subscription-delivery dashboard reads green while subscribers are being dropped.

Suggest gating on the notification shape rather than id-absence, and moving the increment after a successful client write:

```suggestion
			// A backend message carrying a method and no id is a subscription
			// notification pushed by the backend, not a response to a client call.
			// A message WITH an id is a response to a relayed call, already counted
			// at send time in clientPump — counting it here would double-count.
			isNotification = len(res.ID) == 0 && !res.IsError()
```

(with `RecordRPCNotification(w.backend.Name)` moved below the `writeClientConn` success check). Note the ordering also puts this check *before* the `res.IsError()` branch below.

proxyd/metrics.go
```diff
@@ -950,6 +984,90 @@ func boolToFloat64(b bool) float64 {
 	return 0
 }
 
+// statusCodeForRPCRes derives the HTTP-equivalent status code for a single
+// JSON-RPC response, for per-call metrics.
+//
+// An error with HTTPErrorCode == 0 means the backend answered HTTP 200 while
+// carrying a JSON-RPC error — doForward only stamps HTTPErrorCode when the
```
zhwrd: **"doForward only stamps `HTTPErrorCode`" understates the writers — and one of them lets a remote service pick the label value.**

Grep shows four writers, not one: backend.go:975 (doForward), backend.go:302 and backend.go:319 (`ParseInteropError`), and interop_strategy.go:548.

The interop ones matter here. `prepareRPCForForward`'s `txFilter.Apply` branch (server.go:650-653) performs a live RPC to the **interop filter backend**, and `ParseInteropError` copies that filter's raw upstream HTTP status straight through:

```go
fallbackErr.HTTPErrorCode = httpErr.StatusCode
```

`statusCodeForRPCRes` then emits it verbatim. Two effects:

- Those calls are recorded `backend_name="proxyd"` (server.go:729), so an operator paging on `rpc_calls_total{backend_name="proxyd", status_code=~"5.."}` investigates proxyd itself when the interop filter is what degraded.
- The `status_code` label domain is no longer proxyd's fixed error table — it's whatever the filter emits (bounded only by 100-599), multiplied across `method_name × backend_name × transport`. Still bounded, but wider than the comment implies and sourced from a remote service.

Worth either a dedicated sentinel backend name for filter-originated failures, or at minimum correcting the comment so the next reader doesn't rely on the narrower invariant.

proxyd/integration_tests/rpc_calls_test.go
```diff
@@ -0,0 +1,289 @@
...
+	}
+	reqs = append(reqs, NewRPCReq("11", "eth_chainId", nil))
+	// ErrMethodNotWhitelisted.HTTPErrorCode == 403 (backend.go) — asserted below.
+	reqs = append(reqs, NewRPCReq("12", "definitely_not_whitelisted", nil))
+
+	_, statusCode, err := client.SendBatchRPC(reqs...)
+	require.NoError(t, err)
+	require.Equal(t, 200, statusCode)
+
+	require.Equal(t, beforeOK+1, sumRPCCalls(t, okLabels))
```
zhwrd, line 162: **This test asserts 2 of the 12 increments — the "exactly N increments" claim in its own doc comment is untested.**

The comment says it covers "an N-element batch with mixed outcomes produces exactly N increments with the expected status distribution — the claim the SLO rests on." But the only assertions are `eth_chainId` +1 and `method_not_allowed` +1. The 10 `net_version` elements — which are the *entire first minibatch*, the exact shape of the undercount defect the comment names as IMP-1 — are never checked, and `sumRPCCalls` only sums series matching the supplied label subset so neither label set covers them transitively.

A regression that dropped the whole first minibatch's increments passes this test.

```suggestion
	require.Equal(t, beforeOK+1, sumRPCCalls(t, okLabels))
	require.Equal(t, beforeNetVersion+10, sumRPCCalls(t, netVersionLabels))
```

(with `netVersionLabels` mirroring `okLabels` for `method_name: "net_version"`, and a `beforeNetVersion` snapshot alongside the others). A `beforeTotal+12` assertion over the whole batch would pin the headline claim even more directly.

proxyd/backend.go
```diff
@@ -1526,6 +1550,7 @@ func (w *WSProxier) clientPump(ctx context.Context, errC chan error) {
 			)
 			msg = mustMarshalJSON(NewRPCErrorRes(id, err))
 			RecordRPCError(ctx, BackendProxyd, method, err)
+			RecordRPCCall(BackendProxyd, metricLabel, statusCodeForRPCRes(NewRPCErrorRes(id, err)), RPCRequestSourceWS)
```
zhwrd: **The metric's status is derived from a second, independently constructed response — not the one sent to the client.**

Line 1551 already builds `NewRPCErrorRes(id, err)` for the wire; line 1553 builds an identical one just to read `Error.HTTPErrorCode`. Same pattern at 1579/1581 with `NewRPCErrorRes(req.ID, ErrTooManyRequests)`.

They're only coincidentally identical. Edit the error passed to `mustMarshalJSON` and miss the second copy, and the metric silently reports a status the client never received — precisely the failure mode this metric exists to detect, and one no test can catch since both lines are correct in isolation. `writeRPCError` (server.go:1206-1213) already uses the build-once-then-hand-it-downstream shape.

```suggestion
			errRes := NewRPCErrorRes(id, err)
			msg = mustMarshalJSON(errRes)
			RecordRPCError(ctx, BackendProxyd, method, err)
			RecordRPCCall(BackendProxyd, metricLabel, statusCodeForRPCRes(errRes), RPCRequestSourceWS)
```

(the `msg = mustMarshalJSON(...)` on line 1551 folds into this). Minor bonus: drops 1-3 allocations per malformed WS frame, which is the abuse path.

proxyd/backend.go
```diff
@@ -1552,18 +1578,25 @@ func (w *WSProxier) clientPump(ctx context.Context, errC chan error) {
...
 			go w.processClientRPC(ctx, msgType, req, errC)
 			continue
 		}
 
 		RecordRPCForward(ctx, w.backend.Name, req.Method, RPCRequestSourceWS)
+		// Counted at send time: this call's response returns asynchronously in
+		// backendPump with no id correlation back to this request, so its outcome
+		// is unknowable here. Counting at send time guarantees usage completeness
+		// at the cost of an unknown status.
+		RecordRPCCall(w.backend.Name, w.metricMethodName(req.Method), StatusCodeRelayed, RPCRequestSourceWS)
```
zhwrd: **Recorded before `writeBackendConn`, so a call that never reached the backend is still booked as relayed.**

`w.writeBackendConn(msgType, msg)` is at line 1607; if it fails (write deadline, backend socket down) the call is pushed to `errC` and the proxier torn down — but the increment already happened. Since `StatusCodeRelayed` is deliberately non-numeric, it can never match `status_code=~"5.."`, so this failure produces no error signal at all in the new metric.

Bounded impact (one call per connection, since the failure tears down `clientPump`) and the comment does explicitly accept an unknown status for this path — but the fix is a one-line move of the `RecordRPCCall` to after the `writeBackendConn` error check, which makes "relayed" mean *actually relayed*.

proxyd/backend.go
```diff
@@ -1479,6 +1479,18 @@ func (w *WSProxier) Proxy(ctx context.Context) error {
 	return err
 }
 
+// metricMethodName bounds the method_name label cardinality for WS traffic. See
+// Server.metricMethodName — WS uses its own whitelist.
+func (w *WSProxier) metricMethodName(method string) string {
+	if method == "" {
+		return MethodUnknown
+	}
+	if !w.methodWhitelist.Has(method) {
```
zhwrd, line 1513: **This branch is unreachable at both call sites, and the doc comment claims a guarantee it isn't providing.**

Both production call sites (1581 and 1599) are downstream of a successful `prepareClientMsg`, which returns a nil error only after `w.methodWhitelist.Has(req.Method)` passes (backend.go:1744-1746). So `!Has(method)` can never fire here and the function is an identity in production — only `metrics_test.go` exercises the other branches.

That matters because the comment says it "bounds the method_name label cardinality for WS traffic," when the actual bound is `prepareClientMsg`. A future call site added *before* the whitelist gate would trust this function and get no protection.

Related, and the one place the two whitelists are observably different: `shouldHandleWSRPC` (server.go:926-934) admits a request via `txFilter.IsSubmissionMethod` or `txValidationMethods` **without** consulting `rpcMethodMappings`. So a method in `ws_method_whitelist` + a submission method but absent from `rpc_method_mappings` records as its real name / 429 when rejected by the semaphore (line 1581, WS whitelist) and as `method_not_allowed` / 403 when served (server.go:940, `rpcMethodMappings`). Unusual config, but it makes `sum by (method_name)` non-comparable when it happens.

Simplest resolution: delete this helper, pass `req.Method` directly at both sites with a one-line comment pointing at `prepareClientMsg` as the bound.

proxyd/server.go
zhwrd: Fixed in 4edb38c. Confirmed the single-request path (`HandleRPC`, after the `isBatch` block) has no `DeadlineExceeded` branch and falls to `ErrInternal`.

Rather than inline the conditional I pulled it into `deadlineShortCircuitStatus(isBatch bool)` in metrics.go, so the divergence has one definition and a test (`TestDeadlineShortCircuitStatusMatchesHandlerPath`) that fails if either side drifts.

On the integration test: I did not add one. Reaching this branch on the single-request path requires the deadline to expire *before* the minibatch loop — with one element `numBatches == 1`, so the `ctx.Err()` check runs once, before any forward. That means driving it needs the deadline burned inside `prepareRPCForForward`, i.e. a stubbed interop filter round-trip, which is a fair amount of scaffolding for a three-line mapping. The unit test pins the mapping; the gap is that nothing asserts `HandleRPC` still writes 500 there. Flagging rather than hiding it.

proxyd/server.go
zhwrd: Fixed the `statusCodeOK` half in 4edb38c, and added a comment at the increment stating the envelope-level blindness explicitly.

On the outage case: agreed, and it is sharper than what the PR body says. Previously an all-`ErrNoBackends` batch reached the numerator via `writeRPCError`; now it lands in `{status_code="200"}` and the measured 5xx ratio *falls* during exactly the outage you would want it to catch. I have rewritten the rollout section accordingly and moved the per-call SLI alert ahead of the usage-tile repoint so the bake window is not sitting on an unmonitored gap.

Not changing the hardcoded 200 itself — `writeBatchRPCRes` writes `WriteHeader(200)` and the metric is envelope-scoped by construction, so deriving a status from element outcomes would make the metric disagree with the response it is describing. The per-element view is what `proxyd_rpc_calls_total` is for.

proxyd/metrics.go
zhwrd: Fixed in 4edb38c: `HTTPErrorCode: 413`, matching `ErrRequestBodyTooLarge`.

Calling out that **this changes a client-visible status code** — batches over `max_batch_size` were answered HTTP 200 and are now 413. `integration_tests/batching_test.go` asserted the 200, which is how I know nothing else depends on it; that case now asserts 413. Any client treating non-200 as retryable will start retrying a rejection that will never succeed, so it belongs in the release notes.

Also confirmed your read of the sibling rejections: parse error 400, empty batch 400, body too large 413, drain 503. This was the only one in the 2xx bucket.

proxyd/backend.go
zhwrd: Fixed both halves in 4edb38c: gated on `len(res.ID) == 0 && !res.IsError()`, and the `RecordRPCNotification` moved below the `writeClientConn` error check so the counter means delivered.

`RPCRes` has no `Method` field (rpc.go), so a true "has a method, has no id" test would need a second unmarshal of every backend frame on the hot subscription path. The error check is the part that pays for itself — an id-less error frame is the realistic misclassification.

Not doing the `ValidateRPCReq` half. You are right that `prepareClientMsg` never calls it and that an id-less relayed `eth_subscribe` is therefore reachable. But adding validation there rejects requests proxyd currently forwards, which is a client-visible behavior change on the WS relay path and does not belong in a metrics PR. Worth noting the inflation is not directly client-driven either: the client controls the *request* shape, the backend controls the *reply* shape, and geth does not reply at all to an id-less call. Happy to open a follow-up for the validation gap on its own merits.

proxyd/metrics.go
zhwrd: Fixed the comment in 4edb38c — it now names `ParseInteropError` and interop_strategy.go as writers, states that the `status_code` domain is proxyd's table *plus* whatever the interop filter emits (bounded by 100-599, so cardinality holds, but partly sourced from a remote service), and explains why those calls carry `backend_name="proxyd"`: the filter round-trip happens in `prepareRPCForForward` before a backend is selected.

Not adding a sentinel backend name for filter-originated failures. It would split a label domain that dashboards and the SLO denominator are about to be built against, to distinguish a failure mode that has no consumer yet. Better as a follow-up once the per-call SLI is wired up and we can see whether interop-filter 5xx actually shows up as noise in the `proxyd` bucket.

proxyd/integration_tests/rpc_calls_test.go
zhwrd: Fixed in 4edb38c. Added `netVersionLabels` with a `beforeNetVersion+10` assertion, and took the stronger version too — `beforeTotal+12` over `{transport="http"}`, which pins the headline claim directly. Verified the whole-batch assertion is not vacuous: it fails if any single element stops being counted.

proxyd/backend.go
zhwrd: Fixed in 4edb38c, both sites — the malformed-frame path at 1553 and the semaphore-rejection path at 1581 now build `errRes` once and pass it to both `mustMarshalJSON` and `statusCodeForRPCRes`.

proxyd/backend.go
zhwrd: Fixed in 4edb38c — `RecordRPCCall` moved below the `writeBackendConn` error check. The comment now carries both reasons for the placement: after the write so "relayed" means actually relayed, and at send time rather than on the response because `backendPump` cannot correlate the reply back to this request.

proxyd/server.go
zhwrd: Verified and fixed in 4edb38c.

Confirmed the full-override return at backend.go:1131-1133. For that case `backend_name="proxyd"` is actually the right label — proxyd *did* answer it, from its own consensus tracker, and no backend saw the request. So the code was correct and the comment was wrong; the comment now names both empty-`sb` causes.

The partial-override case was a real defect and is fixed. Rather than change `ForwardWithSource`'s signature (this PR deliberately avoids that, and the override count alone would not tell the caller *which* indices), I marked the responses at the single point where overrides are merged back: `OverrideResponses` sets an unexported `servedLocally` flag on each `ov.res`. That one line covers all five `overriddenResponses` creation sites and both the full and partial paths. `handleBatchRPC`'s scatter-back then reads it and attributes those elements to `proxyd` instead of inheriting `forwardedBy`.

The flag is unexported and absent from `rpcResJSON`, so it never reaches the wire and external callers cannot set it. `TestOverrideResponsesMarksServedLocally` covers the merge point; the three-line server branch is not separately covered — a partial-override integration test needs a `consensus_aware` group plus a block tag past the consensus latest, which I did not build out.

proxyd/backend.go
zhwrd: Verified the unreachability claim — both sites are downstream of a successful `prepareClientMsg`, so the collapsing branches are dead in production today. Comment fixed in 4edb38c.

Pushing back on the deletion, though. The premise that "a future call site added before the whitelist gate would trust this function and get no protection" is not right: the helper consults `w.methodWhitelist` itself, so it is a self-contained bound, not a delegation to `prepareClientMsg`. A call site added ahead of the gate that calls this *does* get protected. It is redundant today; it is not inert.

Given the hazard — client-supplied method strings, unauthenticated clients, unbounded Prometheus cardinality — I would rather keep a `Has` lookup on two WS paths than pass `req.Method` raw and rely on a comment surviving the next refactor of the gate. The comment now says outright that the branches are unreachable from `clientPump`, that only tests exercise them, and why it is kept anyway. If you still want it gone, say so and I will delete it.

Separately: verified the `shouldHandleWSRPC` divergence you found. `handleWSRPCInner` re-checks `rpcMethodMappings` and rejects with `ErrMethodNotWhitelisted`, so such a method is never actually served — both paths reject it, and no call is miscounted. What differs is the `method_name` label on each rejection (real name from the semaphore path, `method_not_allowed` from the served path), which makes `sum by (method_name)` non-comparable across them in that config. Documented on the helper rather than changed, since making them agree means either widening `Server.metricMethodName` to consult three sets or narrowing `shouldHandleWSRPC` — both behavior changes beyond this PR.

proxyd/server.go
zhwrd: Resolving: answered by documentation, not a behavior change. The batch envelope genuinely is HTTP 200 per JSON-RPC semantics, so the comment now states that per-element outcomes are invisible to this metric by construction and points at `proxyd_rpc_calls_total` as the per-element view. The one case where the 200 *was* wrong — `ErrTooManyBatchRequests` — is fixed separately in this commit (now 413).

proxyd/backend.go
zhwrd: Resolving: the helper is kept, with the doc comment now stating outright that both clientPump call sites sit downstream of prepareClientMsg so the collapsing branches are unreachable today and only tests exercise them — kept because the bound is self-contained and label cardinality is the failure mode. The ws_method_whitelist vs rpc_method_mappings divergence is documented too, including that `sum by (method_name)` is not comparable across the two rejection paths in that configuration.

sjonpaulbrown (approved):

PR #727, opened 2026-08-27
proxyd: start tx_monitor inclusion clock at request receipt

zhwrd:
## What

The `tx_monitor` inclusion-latency clock now starts at **request receipt** instead of at backend-ack (post-forward success).

### Why

The previous start boundary excluded proxyd-internal handling (rate limiting, validation, batching) and the forward round-trip to the backend — precisely the latency a "tx UX" monitor exists to catch. The measured window was really "backend-ack → this-replica-notices-block," not "submit → inclusion."

The end boundary is intentionally left as poll-observation wall-clock (not the block header timestamp): it reflects when the user can actually perceive inclusion through the same infra, not when the sequencer produced the block.

### Changes

- `TxMonitor.Observe` takes an `at time.Time` (request-arrival), used directly as the ingest timestamp instead of `m.now()`.
- New `ContextKeyReqReceivedAt` + `GetReqReceivedAt(ctx)` getter (falls back to `time.Now()` when unset, so the start is never zero).
- **HTTP**: `populateContext` stamps arrival at ingress.
- **WS**: `clientPump` stamps a per-**message** arrival time and threads it through a per-message ctx (the connection ctx would otherwise yield connection-open time).

### Rejected txns need no eviction

Enrollment stays on the single success hook, so rejects are excluded structurally:
- Validation rejects return a local response and `continue` before forwarding — never observed.
- Backend rejects (nonce-too-low, underpriced, ...) come back as JSON-RPC error results, filtered by the existing `!res[i].IsError()` guard.

The store already dedups on hash and keeps the earliest arrival time, so resubmits preserve the first-seen submission time.

### Not addressed (inherent to a passive v1)

Accepted-then-dropped txns (mempool eviction, same-nonce replacement) still age out as `txmon_timeouts_total` — cannot be popped without receipt polling.

## Tests

- New `TestTxMonitorObserveUsesArrivalTime`, `TestGetReqReceivedAt`; updated `TestTxMonitorObserveNeverBlocks` for the new signature.
- `go vet ./.`, `go build ./...`, and tx-monitor tests pass.

### DIFF
--- a/proxyd/tx_monitor.go
+++ b/proxyd/tx_monitor.go
@@ -130,10 +130,12 @@ func (m *TxMonitor) Shutdown() {
 
 // Observe records a successfully forwarded tx. Non-blocking by construction:
 // a full channel drops the observation and counts it. This is the ONLY method
-// called from the request hot path.
-func (m *TxMonitor) Observe(hash common.Hash, backendGroup string, source string) {
+// called from the request hot path. at is the request-arrival time captured at
+// ingress, so inclusion latency accounts for proxyd-internal handling and the
+// forward round-trip, not just backend-ack-to-block.
+func (m *TxMonitor) Observe(hash common.Hash, backendGroup string, source string, at time.Time) {
 	select {
-	case m.events <- txmonEvent{hash: hash, backendGroup: backendGroup, source: source, at: m.now()}:
+	case m.events <- txmonEvent{hash: hash, backendGroup: backendGroup, source: source, at: at}:
 	default:
 		txmonDroppedChannelFull.Inc()
 	}
--- a/proxyd/CHANGELOG.md
+++ b/proxyd/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Minor Changes
 
-- Added passive tx-UX monitor (`[tx_monitor]`): inclusion-latency histograms from forwarded `eth_sendRawTransaction` to block/subblock visibility.
+- Added passive tx-UX monitor (`[tx_monitor]`): inclusion-latency histograms measured from `eth_sendRawTransaction` request receipt (covering proxyd-internal handling and the forward round-trip) to block/subblock visibility.
 
 ## 3.14.1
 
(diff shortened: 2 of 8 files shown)


proxyd/tx_monitor.go
```diff
@@ -130,10 +130,12 @@ func (m *TxMonitor) Shutdown() {
...
 // a full channel drops the observation and counts it. This is the ONLY method
-// called from the request hot path.
-func (m *TxMonitor) Observe(hash common.Hash, backendGroup string, source string) {
+// called from the request hot path. at is the request-arrival time captured at
+// ingress, so inclusion latency accounts for proxyd-internal handling and the
+// forward round-trip, not just backend-ack-to-block.
+func (m *TxMonitor) Observe(hash common.Hash, backendGroup string, source string, at time.Time) {
 	select {
-	case m.events <- txmonEvent{hash: hash, backendGroup: backendGroup, source: source, at: m.now()}:
+	case m.events <- txmonEvent{hash: hash, backendGroup: backendGroup, source: source, at: at}:
```
pcw109550, line 138: nit; this `at` will be populating `IngestedAt` of `txmonEntry`. Should we take care of https://github.com/ethereum-optimism/infra/blob/04e341f66332d4061a9c1b94d6c7a0b8c43acf8e/proxyd/tx_monitor_store.go#L42-L47

to consider the first observed tx? So something like
```go
if existing, ok := s.txs[e.Hash]; ok {
    if e.IngestedAt.Before(existing.IngestedAt) {
        s.txs[e.Hash] = e
    }
    return true
}
```

pcw109550 (approved):

proxyd/tx_monitor.go
zhwrd: Good catch — done. Since `IngestedAt` is now stamped at request ingress (decoupled from enqueue order), a later-arriving duplicate can be `Observe`d before an earlier one, so `Add` now keeps the entry with the earlier `IngestedAt` (and preserves any already-set `SubblockAt` across the swap). Covered by `TestMemoryPendingStoreAddKeepsEarliest`.

yashvardhan-kukreja (approved):

PR #731, opened 2026-09-02
fix(deps): bump browserslist to 4.28.7

falcorocks:
## Summary

Adds a `browserslist: ^4.28.7` entry to the existing `resolutions` field in `replica-healthcheck/package.json` and regenerates `replica-healthcheck/yarn.lock` with yarn classic. browserslist is a transitive dependency; the lockfile previously contained two entries (4.28.8 and 4.24.4), which now consolidate to a single 4.28.8 resolution.

## Security impact

Resolves the High Dependabot alert GHSA-73wf-gq98-2v4g (alert 375) against browserslist in `replica-healthcheck/yarn.lock`. First patched version is 4.28.7. The two grpc alerts are addressed separately.

## Validation

- `grep` confirms no browserslist entry below 4.28.7 remains in `replica-healthcheck/yarn.lock`.
- The lockfile diff contains no version downgrades: the only version change is browserslist 4.24.4 -> 4.28.8; remaining hunks are range-key consolidation with unchanged resolved versions.
- `yarn install` completes successfully with the updated lockfile.

Part of https://github.com/ethereum-optimism/cloud-security/issues/280

### DIFF
--- a/replica-healthcheck/package.json
+++ b/replica-healthcheck/package.json
@@ -85,6 +85,7 @@
     "ajv": "6.14.0",
     "bn.js": "5.2.3",
     "brace-expansion": "2.1.4",
+    "browserslist": "^4.28.7",
     "diff": "5.2.2",
     "copyfiles/minimatch": "3.1.5",
     "depcheck/js-yaml": "3.15.2",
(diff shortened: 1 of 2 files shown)


hdcesario-op (approved):
LGTM

PR #732, opened 2026-09-02
ci: build apko image twins via the factory apko.yaml workflow

falcorocks:
Add apko-built image twins alongside the existing docker builds, published to the same canonical registry and distinguished by an apko- tag prefix.

- Replace the split apko workflow chain with a matrix caller of the shared factory apko.yaml reusable workflow (one call per service).
- Publish to us-docker.pkg.dev/oplabs-tools-artifacts/images; factory appends the per-service name.
- Every apko tag is apko- prefixed (apko-<sha> on branch pushes, apko-<version> on release tags) so the twins never collide with the docker <sha>/<version> tags in the same repository.
- Inline the local melange pipeline steps into the melange configs.

The docker image build is unchanged.

Closes https://github.com/ethereum-optimism/cloud-security/issues/382


.github/workflows/build-images.apko.yaml
```diff
@@ -6,145 +6,83 @@ on:
...
-      checkout-source: ${{ matrix.checkout_source }}
-      source-submodules: ${{ matrix.source_submodules }}
-      use-github-app-token: ${{ matrix.use_github_app_token }}
-      use-gha-cache: ${{ matrix.use_gha_cache }}
-      cache-key-files: ${{ matrix.cache_key_files }}
+        env:
+          CONFIG: .github/images.json
+        run: |
+          set -euo pipefail
+          MATRIX=$(jq -c '[.apko_images | to_entries[] | .value + {group: .key}]' "$CONFIG")
```
hdcesario-op: This matrix is unconditional, so a service-scoped release tag such as `proxyd/v4.32.2` builds and tags all ten services as `apko-v4.32.2`. That also stamps the proxyd version into unrelated binaries, and a later service release with the same version can overwrite those tags. Could we filter `.apko_images` to the tag’s service for `service/v*` refs, while keeping all entries for global `v*` tags?

.github/images.json
```diff
@@ -60,5 +60,77 @@
       "dockerfile": "replica-healthcheck/Dockerfile",
       "paths": ["replica-healthcheck/**"]
     }
+  },
+  "apko_images": {
+    "cci-stats": {
+      "melange_config": "melange/cci-stats.yaml",
+      "apko_config_dir": "apko",
+      "services": "cci-stats",
+      "smoke_test": "",
```
hdcesario-op, line 69: All ten `smoke_test` values are empty, so the pinned reusable workflow skips its smoke-test step. The removed jobs ran command-aware smoke tests on both amd64 and arm64, including the CircleCI toolchain; green CI now proves build and assembly only. Could we preserve the per-architecture command tests, or extend the factory workflow to support them before removing the old jobs?

hdcesario-op (commented):
Inline review feedback

op-owner-falcorocks (approved):

PR #737, opened 2026-09-08
proxyd: send optional x-api-key header from tx validation middleware

trianglesphere:
**Description**

Add an optional `api_key` to `[tx_validation_middleware]`. When set, the validation client sends it on every request in the `x-api-key` header; when unset, no header is sent and behavior is unchanged.

The key is resolved through ReadFromEnvOrConfig, so it can be supplied as `$ENV_VAR` like other secrets in the config. The startup log records only whether a key is configured, never its value.


ajsutton (approved):
LGTM.

PR #744, opened 2026-09-18
fix(proxyd): expand env vars in consensus_ha_redis.url

parkjeongryul:
**Description**

The top-level `[redis].url` goes through `ReadFromEnvOrConfig`, but `consensus_ha_redis.url` is passed straight to `NewRedisClient`. Same `RedisConfig` type, different treatment — so `url = "$REDIS_URL"` is taken literally in the HA block, `ParseURL` chokes on it, and proxyd dies at startup. Looks like an oversight.

This adds the same call the top-level block already makes. Only `.URL` — the HA tracker doesn't read `.ReadURL`.

**Tests**

Added `TestConsensusHARedisURLFromEnv` plus a `consensus_ha.toml` fixture. It starts a miniredis, exports `REDIS_URL`, and boots proxyd with `consensus_ha_redis.url = "$REDIS_URL"` — same setup `caching_test.go` already uses. Reverting the one-line change makes it fail with `redis: invalid URL scheme:`. Existing `TestConsensus*` still pass.


jelias2 (approved):

jelias2:
This will be available in `v4.32.2`, thanks for the contribution!

